Issue #29, opened 2020-02-17
Update to latest version of TEASER


nllong:
Update to pre-release version 0.7.2

PR #30, opened 2020-02-17
update to latest teaser

nllong:
This PR updates the latest version of teaser (version 0.7.2). This also removes the dependency on the fork of TEASER, we are now pointing to the main teaser repo again.

Addresses: #29

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,11 +1,11 @@
 Change Log
 ==========
 
-Version 0.1
------------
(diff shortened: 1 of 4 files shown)


nllong:
> Thanks. I didn't see any problem. I only have one concern: for teaser.py you changed the file format from ".mat" to ".txt". Why? Or is it because of a change from TEASER side?

Yes, this changed on the TEASER side. Presumably, because the files were not really .mat format.

PR #55, opened 2020-03-31
templating geojson_13buildings into Modelica models

YanfeiNREL:
#### A geojson file with information of 13 buildings was templated into Modelica models: default-modelica and RC-based modelica. For each building, each room model was generated into modelicas. 
They are in the tests/tests_13buildings_modelica/ and tests_tests_13buildings_rc4/, respectively. 

#### Issues need to fix in future: 
(1) clean the geojson files with the same type of keys/parameters, such as: floor_area, floor_height, year_built, etc.... Those 13 buildings don't have consistent keys/parameters. If such information is not available, we need to pre-determine some values to fill in.  
(2) we need to include the building mechanical systems, or terminal units into the templating. 
(3) We also need to include the distribution systems into the templating. 
(4) Probably the ETS systems are also worth included. 
(5) Connections between buildings/terminal units/distribution systems/ets, are also worth exploring through this templating approach.


nllong:
Note that the keys in the GeoJSON are defined in the schema. There shouldn't be any reason to have them be anything else.

Issue #94, opened 2020-05-05
Export fraction latent as Modelica parameter


nllong:
For the moisture balance in TEASER, make the fraction latent variable a parameter.

PR #95, opened 2020-05-05
Fraction Latent as a Parameter

nllong:
The fraction latent of a person was being set directly and not using a modelica parameter.

* Update input parser to support adding new parameters
* Add fraLat with a description when post processing the TEASER output.

* py.test -- all tests should pass. 

#94

### DIFF
--- a/geojson_modelica_translator/modelica/input_parser.py
+++ b/geojson_modelica_translator/modelica/input_parser.py
@@ -316,7 +316,7 @@ class InputParser(object):
     def add_parameter(self, var_type, var_name, value, description):
         """Add a new parameter. Will be prepended to the top of the models list
 
-        :param var_type: string, type of Modelica variable, Real, Integer, String.
+        :param var_type: string, type of Modelica variable, Real, Integer, String, Modelica.SIunits.Area, etc.
         :param var_name: string, name of the variable. Note that this does not check for conflicts.
         :param value: variant, value to set the variable name to.
         :param description: string, description of the parameter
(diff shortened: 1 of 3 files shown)


geojson_modelica_translator/modelica/input_parser.py
```diff
@@ -313,6 +313,22 @@ def add_model_object(self, model_name, model_instance, data):
             str += f"    {d}\n"
         self.model["objects"].append(str)
 
+    def add_parameter(self, var_type, var_name, value, description):
```
nllong, line 316: needed to add a new method. @macintoshpie this will need to be added to modelica builder...

geojson_modelica_translator/modelica/input_parser.py
```diff
@@ -313,6 +313,22 @@ def add_model_object(self, model_name, model_instance, data):
...
+        """Add a new parameter. Will be prepended to the top of the models list
+
+        :param var_type: string, type of Modelica variable, Real, Integer, String.
```
Helarga, line 319: :param var_type: string, type of Modelica variable, Real, Integer, String, Boolean or it could be like this 
   parameter Modelica.SIunits.Area ANor "Area of North wall"
or 
 parameter Modelica.SIunits.Temperature Tind "Indoor Air temperature"

nllong: oh, so it can be a modelica type... okay. I can update the documentation string.

Issue #63, opened 2020-04-10
Replace custom JSON parsing with JsonPath


nllong:
The get methods developed for system parameters uses a custom searching format. This needs to be replaced with an official JsonPath (similar to XPath).

PR #111, opened 2020-05-19
Replace dot delimited strings with JsonPath

nllong:

The original `get_param` method was a custom method to navigate a dictionary by delimiting on periods in a string.

* Use JsonPath as the mechanism to navigate a python dictionary (loaded from a JSON file)

* `py.test` should cover all the cases.

#63


nllong:
@YanfeiNREL can you review this PR... it should be quick, thanks!

PR #138, opened 2020-07-16
Tweaks to enable GMT-Examples repo

nllong:
* Allow larger files in pre-commit
* Update the connection template
* propagate the year built to the teaser ETS file
* rename the Spawn Building ETS python file.


nllong:
ah, yes, you need to update your modelica-builder dependency. I think just running 

`pip install -U --upgrade-strategy eager -r requirements.txt`

will work.

vtnate:
Yep, that did the trick. Unfortunately, just like in #146, py.test is also failing on this branch when running `DistrictCoolingSystem.mo` from `tests/model_connectors/test_district_system.py::SpawnModelConnectorSingleBuildingTimeSeriesTest::test_district_to_modelica_and_run`. Could it be that the new MBL dependency wants a reformatted *.mo file, a la #137 ? OTOH, the CI check including that same test passed. 🤔

nllong:
> Yep, that did the trick. Unfortunately, just like in #146, py.test is also failing on this branch when running `DistrictCoolingSystem.mo` from `tests/model_connectors/test_district_system.py::SpawnModelConnectorSingleBuildingTimeSeriesTest::test_district_to_modelica_and_run`. Could it be that the new MBL dependency wants a reformatted *.mo file, a la #137 ? OTOH, the CI check including that same test passed. 🤔

I'm going to merge since travis is happy. We need to figure out what is going on with your machine though.

PR #148, opened 2020-07-24
Amys work 136--updates to setup.py script for Windows

amyeallen1:
I made some updates to the setup.py script for Windows to deal with problems involving too-long file paths, and uncommented the specific SHA, since without that I ran into a hiccup involving a package.order file. Nick, @nllong, could you please review when you have a chance? Thanks.


setup.py
```diff
@@ -86,20 +87,31 @@
...
-            if os.path.join(lib_to_extract.replace("/", os.path.sep)) in file:
-                print(f"extracting ... {file}")
-                zip.extract(file, path=tmp_save_path)
+            if platform == "win64" or platform == "win32": ##AA added this to resolve an error due to a too-long file path 
+                #if os.path.join(lib_to_extract.replace("/", os.path.sep)) in file:
+                if os.path.join(lib_to_extract) in file:
+                    print(f"extracting ... {file}")
+                    #zip.extract(file, path=tmp_save_path)
+                    zip.extract(file) ##AA replaced this due to file path length 
+                else: #AA added, for non-Windows case
```
vtnate, line 105: Should this `else` statement be one indent farther out?

nllong: I think it does. But I think we can actually remove the checkout of MBL for now since we are currently requiring the user to have their own MODELICAPATH set correctly.

amyeallen1: Yes, it should be. Sorry about that! @vtnate @nllong Nick, should I remove the MBL checkout and re-commit?

nllong: Just documenting our slack communication. I will cherry pick your changes and fix this in develop.

amyeallen1: @nllong Okay, great, thanks, Nick!

PR #160, opened 2020-08-03
set output interval length to 3600S for the timeSeries template.

Helarga:
The output interval length for the time series models have been modified to 3600S.

Before 
image

After 
image


geojson_modelica_translator/model_connectors/templates/TimeSeriesCouplingBuilding.mot
```diff
@@ -83,6 +83,7 @@ annotation (Diagram(
          "Simulate and plot"),
         experiment(
             StopTime=604800,
+            Interval=3600,
```
nllong, line 86: 👍

PR #172, opened 2020-08-20
Proposed updates to schema

amyeallen1:
Hi Nick and Hagar, this PR has the proposed changes to the schema that we discussed last week. The list of ETS properties could be extended in the future as we support more ETS types. The "ets_name" property is a placeholder to allow one to set different values of properties for heating vs. cooling. @Helarga @nllong


amyeallen1:
Hi Hagar and Nick, 
Hagar, thanks for reviewing this. I made the changes you suggested (making the units mass flow rate as opposed to volumetric flow rate) and removed some duplicate content. @Helarga @nllong

geojson_modelica_translator/system_parameters/schema.json
```diff
@@ -25,6 +25,101 @@
           },
           "additionalProperties": false
         },
+        "ets": {
```
nllong, line 28: with this setup it appers that all buildings will get the same ETS system, is that what is expected?

geojson_modelica_translator/system_parameters/schema.json
```diff
@@ -25,6 +25,101 @@
...
+              },
+              "additionalProperties": {
+                "type": "string"
+              }
+            },
+            "ets_type_heating": {
+              "description": "Energy transfer station model for heating, one side is connected with district water loops, and the other side is connected with building water loops",
+              "type": "string",
+              "enum": [
+                "1G Indirect",
```
nllong, line 96: I'm not sure I like the 1G, 2/3 G style naming. Any other thoughts on making this look better? Maybe break up generation as (First Generation, Second Generation), and ETS connection type (Direct, Indirect)?

geojson_modelica_translator/system_parameters/schema.json
```diff
@@ -25,6 +25,101 @@
...
+                "1G Direct",
+                "2/3 G Direct",
+                "2/3 G Indirect",
+                "4/5 G Direct",
+                "4/5 G Indirect"
+              ]
+            },
+            "ets_type_cooling": {
```
nllong, line 104: I renamed this to ets_type_cooling instead of ets_type_clg.

tests/system_parameters/data/system_params_1.json
```diff
@@ -79,6 +51,20 @@
       }
     ]
   },
+  "ets": {
```
nllong, line 54: Can you update the various system_params JSON files in the repo to the new schema format? This is in part why the tests are failing.

nllong (changes requested):
This is great, thanks for updating the schema. I have a few requested changes. 

Make sure to run `py.test` locally to ensure all the tests pass.

geojson_modelica_translator/system_parameters/schema.json
amyeallen1: Good point! I revised the structure of the schema so that one can set defaults overall, and then customize at the building level.

geojson_modelica_translator/system_parameters/schema.json
amyeallen1: That makes sense. I revised the naming, and added another property so that the generation and connection type are separate properties.

tests/system_parameters/data/system_params_1.json
amyeallen1: Good point. I updated the test files.

nllong (approved):

nllong (approved):

amyeallen1:
Thanks, Nick!

Issue #170, opened 2020-08-19
[8/28] refactor to have `add_building` method in base.py


vtnate:
### Expected Behavior

Add a generic `add_building` method to base.py, to replace the add_building method in each model_connector class.

PR #174, opened 2020-08-25
Add building in base

vtnate:
First step in refactoring gmt
Removes `add_building` method from the spawn, teaser, & time_series model connectors, and consolidates into the base class. Base class is already the super of each of those model connectors, so this was a very simple non-breaking DRY-up.
`py.test tests/model_connectors/test_spawn.py` (or any of the model connector tests, or just py.test for everything)
#170

### DIFF
--- a/geojson_modelica_translator/model_connectors/base.py
+++ b/geojson_modelica_translator/model_connectors/base.py
@@ -83,11 +83,11 @@ class Base(object):
 
             try:
                 urbanopt_building.feature.properties["year_built"]
+                # UO SDK defaults to current year, however TEASER only supports up to Year 2015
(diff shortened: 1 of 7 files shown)


geojson_modelica_translator/model_connectors/base.py
```diff
@@ -60,6 +60,48 @@ def __init__(self, system_parameters):
         self.required_mo_files = []
         # extract data out of the urbanopt_building object and store into the base object
 
+    def add_building(self, urbanopt_building, mapper=None):
```
nllong, line 63: thanks, this looks much better!

nllong (approved):

Issue #169, opened 2020-08-19
[9/25] Refactor Scaffold to support holding all files/objects for project


nllong:
Schema updates will conflict. Waiting for Amy's PR. 

Presently the scaffold holds only the name of the directions. The scaffold should be updated to hold instances of buildings, district networks, plants, etc.

PR #180, opened 2020-09-14
[refactor/scaffold] Move model connectors to lookup

nllong:
The to_modeilca method was too rigid and only handled a single load model type (e.g., spawn or time series) at a time.

* Support multiple load models at a time
* Rename load types to [spawn, rc, and time_series] and update schema

`py.test`

#169 
There will be other updated includes
  * abstract out an isolated building vs ETS building to remove redundant code
  * move district test to place loads in the correct location, create multiple building loads as defined in geojson


vtnate (approved):
I love it when a PR has more deleted lines than added. 😁

nllong:
@amyeallen1 and @vtnate  -- I'm going to hold off merging this down as it will break your various branches.

amyeallen1:
Thanks, Nick!

vtnate:
@nllong Did #184 break this? Lines 327-328 in the Travis output make it look like it's not my fault, but 329-330 have my fingerprints on them... 😟

nllong:
Well, yes it did, but that is only because I have changed the assumption that a `model_connector` is for only one load/building now. I am fixing it.

PR #183, opened 2020-09-16
Single time_series.py model for ETS and w/o ETS

nllong:

The load connectors (model connectors) have multiple files to denote if the building has an ETS or not. That "flag" of whether or not the ETS exists should come from the system parameters file.

* A single time_series.py file that has two paths through for No ETS (resulting in coupling.mo) and with an ETS (resulting in TimeSeriesETSCoupling.mo).
* Added new base test helper method for running docker command and doing the basic checks (exitcode = 0)
* Started implementing the assumption that a LoadConnector is only a single building. This will greatly simplify things as the connectors will only know about a single building and the other classes will know of the multiple buildings. This also illustrates the issue that the package.order files should not be written by the connector, but rather a global scaffold method (or similar).
* Note that multiple buildings/loads should only be tested from the main GMT to_modelica method. The model_connectors tests should only test single buildings.

`py.test` 

#169


tests/model_connectors/test_time_series.py
```diff
@@ -29,30 +29,24 @@
...
     TimeSeriesConnector
 )
-from geojson_modelica_translator.modelica.modelica_runner import ModelicaRunner
 from geojson_modelica_translator.system_parameters.system_parameters import (
     SystemParameters
 )
 
+from ..base_test_case import TestCaseBase
 
-class SpawnModelConnectorSingleBuildingTimeSeriesTest(unittest.TestCase):
```
nllong, line 48: ugh... copy-pasting is the worst, and I know that I'm the culprit.

nllong:
@vtnate --can you check this one too. This is getting merged into the `169-scaffold` branch.

tests/base_test_case.py
```diff
@@ -60,3 +63,25 @@ def set_up(self, root_folder, project_name):
             shutil.rmtree(os.path.join(self.output_dir, project_name))
 
         return self.data_dir, self.output_dir
+
+    def run_and_assert_in_docker(self, file_to_run, project_path, project_name):
+        """
```
vtnate, line 68: This makes me so happy

geojson_modelica_translator/model_connectors/time_series.py
```diff
@@ -47,79 +47,147 @@ def to_modelica(self, scaffold):
...
+                os.path.join(b_modelica_path.scripts_dir, "RunTimeSeriesBuilding.most"),
+                full_model_name=os.path.join(
+                    scaffold.project_name,
+                    scaffold.loads_path.files_relative_dir,
+                    f"B{building['building_id']}",
+                    "coupling").replace(os.path.sep, '.')
+            )
+        # TODO: Need to update this to include items other than Indirect Cooling, should just be indirect.
+        elif ets_model_type == "Indirect Cooling":
+            # The ETS model is Indirect Cooling. If this model is to be connected to a district, then
```
vtnate, line 149: Yay!

vtnate (approved):
All green, LGTM

geojson_modelica_translator/model_connectors/time_series.py
nllong: ha! yeah, that comment was specifically for your work.

PR #187, opened 2020-09-30
Use modelica builder from pypi

nllong:
We just released the first version of Modelica Builder and should use that version instead of the git version.

* Use Modelica-Builder from PyPI (0.1.0)
* Upgrade testing dependencies, requests, and jsonpath-ng. Others can be upgraded but they seemed too risky before release.

* `pip install -r requirements.txt`
* Verify that tests still run and pass with new dependency.

n/a

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -161,15 +161,38 @@ Templating Diagram
 Release Instructions
 --------------------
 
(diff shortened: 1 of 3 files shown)


vtnate (approved):
Downright official! 🐮 🐴 🐑 🧑‍🌾

PR #189, opened 2020-09-30
Do not pass system_params

nllong:
The process_loads() method required the passing of the system parameters object.

* Remove the system_params from the process_loads. It is not needed because the geojson object already has the system_parameters file as a member variable. It is set by calling gj.set_system_parameters(<a_file_obj>)

* `py.test` should cover the tests

N/A


vtnate (approved):
Everything passed locally too. Yay for simplification.

PR #213, opened 2020-10-30
Update TimeSeriesCouplingETS.mot

Helarga:
assign the time series's distance between output points "output intervals" to 3600 seconds.


nllong (approved):
thanks!

PR #223, opened 2020-11-03
refactor/time-series-ets-poc

macintoshpie:
This is a proof of concept for how GMT could generate and couple models.
Demonstrates how TimeSeries load and Cooling Indirect can be generalized for generic coupling between systems, rather than having to explicitly define a higher level modelica file that connects models.

Here's a diagram of the processing flow.
Screen Shot 2020-11-04 at 9 02 05 AM

Generally, we now have the concept of "couplings" which is a pairing of two systems. Since the requirements of each pairing is unique, we have to define separate templates for each possible pair, e.g. TimeSeries and CoolingIndirect.

Each coupling is defined as a set of template files:
1. First is a template for defining variables that are necessary for the pairing, e.g. if model A requires some value calculated from model B. This file is `<ModelA>_<ModelB>/ComponentDefinitions.mopt` (note that the ordering of the directory name depends on the two systems being linked, with priority, in order, being: load, ets, network.)
2. Second is a template for defining connect statements between the models. This file is `<ModelA>_<ModelB>/ConnectStatements.mopt`

Couplings are defined as objects using the Coupling class, then given to a District model, which will render these templates, and insert them into its **own** template to represent a higher level model, ie a district.

In addition, each model type (TimeSeries, CoolingIndirect, etc) must define an _instance_ template, which is the component definition to be inserted into the district (this is not handled by the coupling). This is the instantiation of the .mo file we generate for the model elsewhere.

As an example, lets consider that we have a TimeSeries load and Cooling Indirect ets we want to model. The required template files are:
```
model_connectors/
  - couplings/
    - templates/
      - TimeSeries_CoolingIndirect/
        - ComponentDefinitions.mopt
        - ConnectStatements.mopt
  - templates/
    - TimeSeries_Instance.mopt (note that this and the CoolingIndirect instance will be moved into their own directories as we refactor the organization of templates)
    - CoolingIndirect_Instance.mopt
    - TimeSeriesBuilding.mot (this is the traditional mot file we use to generate our "base" model)
    - CoolingIndirect.mot (also the traditional mot file we use to generate our base model)
```

If we create a coupling of TimeSeries and CoolingIndirect, the district model will do the following:
1. call `to_modelica` for each model (this is the method which generates the .mo file for the model)
2. render the partial templates for the coupling (component definitions and connect statements)
3. render the instances of the models
4. insert rendered partials into DistrictEnergySystem.mot

Refer to the notes in developer_resources for more information.

Run the test, the inspect the model output of `test_district_system_new.py` by loading it into Dymola.


nllong:
I just updated the required checks... everything is passing. I'll review in a bit.

nllong (approved):
This looks great. I think we will want to wait until the walkthru with the broader team before we merge this down.

Nicely done Ted!

nllong:
@macintoshpie -- I'm thinking we should merge this down and start tackling the other model flexibilities in new PRs (and issues). Thoughts?

macintoshpie:
I agree, merging now and will get started on the other models

PR #235, opened 2020-11-24
refactor: set default temperature values for spawn load

macintoshpie:
Moves parameters that we don't change into Spawn model itself. Eventually these will be templated here inside the model.
Please run the test for test_spawn_new.py, then load the resulting package into Dymola, and make sure it runs correctly.


nllong (approved):
noice... thanks for moving this out of the coupling.

PR #236, opened 2020-11-25
feat: coupling graph

macintoshpie:
Current and future challenges of GMT necessitate a centralized graph representation of connected systems.
Creates a class to represent this graph of systems.
There was no change to actual behavior, just refactored into a new class. If you really want to, run the *_new.py tests and load them into Dymola

### DIFF
--- a/geojson_modelica_translator/model_connectors/couplings/graph.py
+++ b/geojson_modelica_translator/model_connectors/couplings/graph.py
@@ -1,0 +1,83 @@
+"""
+****************************************************************************************************
+:copyright (c) 2019-2020 URBANopt, Alliance for Sustainable Energy, LLC, and other contributors.
+
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without modification, are permitted
+provided that the following conditions are met:
+
+Redistributions of source code must retain the above copyright notice, this list of conditions
+and the following disclaimer.
+
+Redistributions in binary form must reproduce the above copyright notice, this list of conditions
+and the following disclaimer in the documentation and/or other materials provided with the
+distribution.
+
+Neither the name of the copyright holder nor the names of its contributors may be used to endorse
+or promote products derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR
+IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
+CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT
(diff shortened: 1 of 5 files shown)


geojson_modelica_translator/model_connectors/couplings/graph.py
```diff
@@ -0,0 +1,83 @@
...
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT
+OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+****************************************************************************************************
+"""
+from collections import defaultdict
+
+
+class CouplingGraph:
```
macintoshpie, line 33: I'm not sure what the name of this class should be, open to feedback

geojson_modelica_translator/model_connectors/couplings/graph.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @property
+    def couplings(self):
+        return [coupling for coupling in self._couplings]
+
+    @property
+    def models(self):
+        return [model for _, model in self._models_by_id.items()]
+
+    def couplings_by_type(self, model):
```
macintoshpie, line 61: as a note this method will be refactored soon to support multiple connections to the same type. E.g. it can't currently handle a network that's connected to multiple ETSes.

geojson_modelica_translator/model_connectors/districts/district.py
```diff
@@ -50,31 +49,15 @@ def render_template(template_name, template_params):
     return template.render(template_params)
 
 
-class District(object):
```
macintoshpie, line 53: It's not necessary to subclass object anymore in python 3
https://stackoverflow.com/questions/54867/what-is-the-difference-between-old-style-and-new-style-classes-in-python

nllong: woah. noice. Didn't realize this.

geojson_modelica_translator/model_connectors/couplings/graph.py
nllong: I like the name you have, `CouplingGraph`.

geojson_modelica_translator/model_connectors/couplings/graph.py
nllong: 👍

tests/model_connectors/test_district_system_new.py
```diff
@@ -84,15 +87,17 @@ def test_district_system(self):
...
         district = District(
             root_dir=self.output_dir,
             project_name=project_name,
             system_parameters=sys_params,
-            couplings=[
-                ts_ci_coupling,
-                ci_cw_coupling,
-                ts_hw_coupling,
-            ]
+            coupling_graph=graph
```
nllong, line 100: nice

nllong (approved):
Looks like a great start at making this more flexible. I ran the tests, inspected the files, and ran the teaser version in Dymola.

amyeallen1:
Thanks for doing all this, Ted! I'm running into some hiccups in running the tests, so I haven't been able to generate all of the output files, but I'll talk to Nick about troubleshooting those. I think I understand the approach. Thinking about a future use case-- could the coupling approach between a network and an ETS support the connection of an ETS to a particular port on the network? If I'm understanding correctly, it looks like one could accomplish this now by putting loads in a specific order in the GeoJSON file. 
@macintoshpie @nllong

macintoshpie:
Great question @amyeallen1 , right now there's no proper way to specify which port an ETS should connect to for a network. I think you're right that it should be dependent on the GeoJSON/system parameters files. I'm not sure if we would want to just use order or come up with a way for describing graphs in the systems parameters file.

I'm also wondering if it'll ever be necessary for a user to specify which port they want to connect an ETS on, or if that should all be determined internally by the GMT. I think some examples of why you'd want to connect an ETS to a specific port would be really helpful.

amyeallen1:
Thanks, Ted! I agree that for this use case, it would be preferable to have the the schema for the GeoJSON file able to support a graph or another structure to "match" a building to a specific physical location in the network, where a corresponding port is located. This would be of interest to represent the pressure drop and heat loss that are influenced by the length and characteristics of the piping network. 
@macintoshpie

Issue #240, opened 2020-12-01
revise the schema.json


Helarga:
The schema file needed to be revised to make sure all the design parameters are matching the modelica files

https://github.com/urbanopt/geojson-modelica-translator/blob/develop/geojson_modelica_translator/system_parameters/schema.json

PR #241, opened 2020-12-02
modification of the schema.json 

Helarga:
The system parameter schema.json needed to be reviewed and updated to reflect the latest modelica models.

The PR addresses issue https://github.com/urbanopt/geojson-modelica-translator/issues/240 where 

- Buildings parameters have been updated to include T heating/cooling air set_point, Chilled/heating water supply and return temperature for building and district level.

- Removed  a non required old parameter section" central plant".


geojson_modelica_translator/system_parameters/schema.json
```diff
@@ -287,91 +269,137 @@
...
-          "default": 999
+          "default": 500
         },
-        "cooling_supply_water_temperature_district": {
-          "description": "Cooling supply water temperature for district (deg C)",
+        "TDis_ChWS": {
+          "description": "Chilled water supply temperature district side (K)",
           "type": "number",
-          "default": 5
+          "default": 5+273.15
```
vtnate, line 313: I don't think json allows the `+` symbol in a value. May need to make these default values just a number. For instance, this one would be `278.15`.

geojson_modelica_translator/system_parameters/schema.json
```diff
@@ -199,7 +199,7 @@
           "enum": [
             "None",
             "Indirect Heating and Cooling",
-            "Heat Exchanger Only"
+            "Direct Heating and Cooling"
```
vtnate, line 202: Python code will have to be changed to match these new values you're setting. I can work on that if that will be helpful to you.

geojson_modelica_translator/system_parameters/schema.json
Helarga: ok i will fix it

geojson_modelica_translator/system_parameters/schema.json
Helarga: the "Heat Exchanger only" type is a mistake actually, i corrected it by "Direct Heating and Cooling"

nllong:
Move to urbanopt organization... track here: https://github.com/urbanopt/geojson-modelica-translator/pull/274

Issue #230, opened 2020-11-23
District system to building load flexibility refactor


nllong:
Create an end-to-end district system using the new flexibility components.

PR #244, opened 2020-12-03
feat: network and heating plant

macintoshpie:
Part of improved GMT flexibility effort. We've been working towards representing a full district system.
Implements the network and heating plant, and connects them to our existing systems.
Run the test `test_district_heating_district_new.py`. Inspect the resulting output, there should now be files in the `Plants` direcrtory, and the DistrictEnergySystem.mo should look right. Load it into Dymola and run the simulation.
#230 
Believe it or not, this is _after_ trying to reorganize the generated components to make it look pretty
Screen Shot 2020-12-02 at 5 04 20 PM


nllong:
EEK---sorry for the conflicts @macintoshpie

macintoshpie:
@Helarga It looks like the jmodelica is still not working properly after making the changes suggested for the 2 load model (changes made here 82b1434). It ran for 6 hours and didn't finish here: https://github.com/urbanopt/geojson-modelica-translator/pull/244/checks?check_run_id=1501958206#step:6:50

Could you try running the most recent version of this branch and tell me your thoughts on what we should do next?

geojson_modelica_translator/model_connectors/templates/TimeSeriesBuildingNew.mot
```diff
@@ -1,225 +1,303 @@
-within {{project_name}}.Loads.{{model_name}};
-model building
+within {{ project_name }}.Loads.{{ model_name }};
```
macintoshpie, line 1: ## NOTE
1. This change introduces a new TimeSeries model. It is not fully compatible with the previous model, so certain attributes/properties will be different and need to be handled. This means tests that use this model are going to fail until we get them updated.
2. This new model does *no templating*, meaning the geojson and sys params have no affect (effect??) currently. This will need to be updated accordingly

.github/workflows/ci.yml
```diff
@@ -53,6 +53,13 @@ jobs:
             echo "$filename"
             cat "$filename"
           done
+      -
```
nllong, line 56: Thanks for adding this!

nllong (approved):

PR #246, opened 2020-12-03
Time series heating indirect modifications

Helarga:
Updated the heating indirect ETS parameters and BC " replaced the cooling with the heating".

@macintoshpie so far it is good, kindly merge my branch to feat/time-series-heating-indirect


macintoshpie (approved):
Looks great, thanks for fixing this for me!

PR #251, opened 2020-12-14
fix: use correct value for iConDpSen which is number of connected ETSes

macintoshpie:
Previous instance of our network pipe had a hardcoded value of 2 for `iConDpSen` which would result in an out of bounds exception in the model compilation/simulation when only connecting a single ETS.

Uses the correct number based on number of connected ETSes

Dymola process if you feel its necessary

### DIFF
--- a/geojson_modelica_translator/model_connectors/templates/Network2Pipe_Instance.mopt
+++ b/geojson_modelica_translator/model_connectors/templates/Network2Pipe_Instance.mopt
@@ -25,7 +25,7 @@
   {{ model.modelica_type }} {{ model.id }}(
     redeclare final package Medium={{ globals.medium_w }},
     final nCon=nBui_{{ model.id }},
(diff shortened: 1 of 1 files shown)


Helarga (approved):

geojson_modelica_translator/model_connectors/templates/Network2Pipe_Instance.mopt
```diff
@@ -25,7 +25,7 @@
   {{ model.modelica_type }} {{ model.id }}(
     redeclare final package Medium={{ globals.medium_w }},
     final nCon=nBui_{{ model.id }},
-    iConDpSen=2,
+    iConDpSen=nBui_{{ model.id }},
```
nllong, line 28: noice fix!

nllong (approved):

Issue #254, opened 2020-12-14
DES CLI


laurenklun:
working on going from UO-SDK outputs to be a UO-DES input

PR #250, opened 2020-12-11
Write sys params

vtnate:
#### Any background context you want to provide?
First connection between UO SDK & UO DES
#### What does this PR accomplish?
Reads files from UO SDK and writes a system_parameters.json file for use in DES
#### How should this be manually tested?
`py.test tests/geojson/test_csv_to_sys_param.py`
~User/tester needs to run a scenario with the UO SDK first, and edit line 42 of `test_csv_to_sys_param.py` to point to the scenario folder once the run is complete. SDK run needs to include [Amy's measure(s)](https://github.com/urbanopt/DES_HVAC/tree/develop/Measures), which aren't in the SDK yet, which adds another wrinkle to manual testing.~
#### What are the relevant tickets?
#211 
#### Screenshots (if appropriate)

@nllong I've got a couple TODO & FIXME's in here, with further work that needs to be done. This shows my current progress, and shouldn't be too crazy to move into the SystemParameters class if that's where we want this to live.

PR #256, opened 2020-12-14
Cli

vtnate:
- This is branched off the `write-sys-params` branch, so it includes updates from #250. That PR should be merged first.
- The uo_des CLI currently assumes the user has already run simulations with the uo sdk CLI.
    - Perhaps at some point in the future one cli will become the primary and call the other cli to accomplish tasks.
Creates a user interface for the GMT. At this time only running the CSVToSysParam class
- `pip install .` (to install the new cli from setup.py)
- `uo_des --help`

Expect to see help text with list of available cli commands, currently only `sys-param`
For further details, help text for the command can be seen with `uo_des sys-param --help`
#254


management/uo_des.py
```diff
@@ -0,0 +1,49 @@
...
+    '--sys_param_filename',
+    default="test_sys_param.json",
+    help='Name of systems parameters file to create. Include ".json" in your name',
+    required=True
+)
+@click.option(
+    "-s",
+    "--scenario_dir",
+    help="Path to an urbanopt scenario, probably in the 'run' folder of your UO SDK project directory",
+    default=Path.cwd() / "tests" / "geojson" / "data" / "sdk_output_skeleton" / "run" / "baseline_15min",
```
macintoshpie, line 24: I might remove the default for all options except for the sys_param_filename.

management/uo_des.py
```diff
@@ -0,0 +1,49 @@
...
+)
+@click.option(
+    "-t",
+    "--sys_param_template",
+    default=Path.cwd() / "tests" / "geojson" / "data" / "time_series_template.json",
+    help="Path to the template sys_param.json file",
+    required=True
+)
+def sys_param(sys_param_filename, scenario_dir, sys_param_template, feature_file):
+    """Test GMT CLI."""
```
macintoshpie, line 42:
```suggestion
    """GMT CLI."""
```
this is for realsies right?

management/uo_des.py
```diff
@@ -0,0 +1,49 @@
...
+    # click.echo(sys_param_filename)
```
macintoshpie, line 43:
```suggestion
```

management/uo_des.py
```diff
@@ -0,0 +1,49 @@
...
+    required=True
+)
+@click.option(
+    "-t",
+    "--sys_param_template",
+    default=Path.cwd() / "tests" / "geojson" / "data" / "time_series_template.json",
+    help="Path to the template sys_param.json file",
+    required=True
+)
+def sys_param(sys_param_filename, scenario_dir, sys_param_template, feature_file):
```
macintoshpie, line 41: might be nice to have an option for the `overwrite` parameter for `csv_to_sys_param` call. Also, I think it should default to `False` since we already provide a default sys params filename. But maybe it's not likely users would rerun from a different UO SDK project? Yalls call really

macintoshpie (commented):
A few small comments, but this works great. I ran it locally and it produced the expected json file.

click is pretty nice, eh @nllong 😉

management/uo_des.py
```diff
@@ -0,0 +1,49 @@
...
+from geojson_modelica_translator.geojson.csv_to_sys_param import CSVToSysParam
+
+
+@click.group()
+def cli():
+    pass
+
+
+@cli.command()
+@click.option(
```
nllong, line 13: oh click :p

management/uo_des.py
nllong: let's wait until this use case emerged.

management/uo_des.py
nllong: does the unit tests require the defaults to be set? But yes, I agree with Ted.

setup.py
```diff
@@ -62,12 +62,14 @@
...
     license=license,
     packages=find_packages(exclude=("tests", "docs")),
+    include_package_data=True,
     cmdclass={
         "update_schemas": UpdateSchemas,
         "update_licenses": UpdateLicenses,
         "format_modelica_files": FormatModelicaFiles,
     },
     install_requires=[
+        "Click==7.1.2",
```
nllong, line 72: 👍

tests/geojson/data/sdk_output_skeleton/run/baseline_15min/8/015_export_modelica_loads/modelica.mos
```diff
@@ -0,0 +1,33 @@
+#1
+#Heating and Cooling Model loads from OpenStudio Prototype Buildings
+#  Building Type: {{BUILDINGTYPE}}
```
nllong, line 3: looks like we need to update the export_modelica_loads measure to set the`{{.*}}` lines. Fine for now, just something to clean up later.

nllong (approved):
I suggest that we merge this down and then address the default click args. @macintoshpie can you create an issue and assign to nate?

macintoshpie:
@nllong  Created issue #271 to track updating of cli parameters

nllong:
> @nllong Created issue #271 to track updating of cli parameters

great, thanks. Merging once the tests fail 🙃

PR #260, opened 2020-12-15
Cooling district package

Helarga:
modified templates of central cooling plant and building time series.


macintoshpie (approved):
Looks great, thanks!

PR #266, opened 2020-12-22
refactor: remove ts instance filNam override

macintoshpie:
Current ts building does not use sys params ts load filepath

Uses user-provided ts data

Run test_district_heating_and_cooling_systems.py


macintoshpie:
@nllong Our combined heating and cooling test fails now, see the log here: https://gist.github.com/macintoshpie/fd0beea54e91c5d37f52b543e520cd80

Note that it uses `time_series_large_office_2013_5B.mos`

nllong (approved):
🤘 - nice catch!

PR #273, opened 2020-12-31
Move to sys params

### LINKED ISSUE #271: Update options for uo_des CLI
### Tasks
There shouldn't be defaults for
- scenario_dir
- feature_file
- sys_param_template

Consider updating the default for sys_param to `uo_des_sys_params.json` or something like that rather than `test_...`
Consider adding an option for the `overwrite` parameter.

### LINKED ISSUE #272: Move methods from CSVToSysParam class into SystemParameters class
Enables using jsonpath, and reduces number of classes hanging around.

vtnate:
- Moves creating a sys_param file from uo_sdk output into the SystemParameters class
- Modifies sys_param_template to pass json validation
- Adds `overwrite` flag to cli
- Removes default paths to test files in cli
Automatic testing:
- `py.test tests/system_parameters/test_system_parameters.py`

Using the CLI for realzies
- `uo_des sys-param -p zzz.json -s tests/system_parameters/data/sdk_output_skeleton/run/baseline_15min -f tests/system_parameters/data/sdk_output_skeleton/example_project.json`
Resolves #271 Resolves #272


geojson_modelica_translator/system_parameters/system_parameters.py
```diff
@@ -204,3 +207,81 @@ def validate(self):
...
+            raise Exception(f"Unable to find your feature file. The path you provided was: {feature_file}")
+
+        if Path(sys_param_filename).exists() and not overwrite:
+            raise Exception(f"Output file already exists and overwrite is False: {sys_param_filename}")
+
+        # If template is passed to the class for initialization, it gets automatically validated first.
+        if self.data:
+            param_template = self.data
+        else:
+            raise Exception("No template found. You must pass the sys_param_template location when initializing an instance.")
```
macintoshpie, line 233: I'm not sure what we could do instead, but I feel like it might be a bit confusing for a user if we mention `sys_param_template` here when it's not mentioned elsewhere in this class

geojson_modelica_translator/system_parameters/system_parameters.py
```diff
@@ -204,3 +207,81 @@ def validate(self):
...
+
+        self.measure_list = []
```
macintoshpie, line 235: It might be cleaner to just use a local variable here instead of assigning to self now that we're no longer calling other methods to append to it.

management/uo_des.py
```diff
@@ -13,36 +15,35 @@ def cli():
...
+    '-o',
+    '--overwrite',
+    is_flag=True,
+    help="If a file already exists with the sys_param_filename, overwrite it."
 )
-def sys_param(sys_param_filename, scenario_dir, sys_param_template, feature_file):
+def sys_param(sys_param_filename, scenario_dir, feature_file, overwrite=False):
     """GMT CLI."""
-    csv_to_sys_param = CSVToSysParam(
+    sys_param_template_path = Path(__file__).parent.parent / 'geojson_modelica_translator' / \
```
macintoshpie, line 41: Is this template file always going to be the same (ie we don't allow changing the template)?

If so, I wonder if we should just load this inside of the `csv_to_sys_param` method. That would prevent users from initializing the SystemParameters with a json file that's not what we expect when templating. We could also make `csv_to_sys_param` a class method so the caller doesn't have to instantiate the class themselves (see the `loadd` method).

geojson_modelica_translator/system_parameters/system_parameters.py
```diff
@@ -204,3 +207,81 @@ def validate(self):
...
+
+        # Remove template buildings that weren't used or don't have successful simulations, with modelica outputs
+        building_list = [x for x in building_list if not x['load_model_parameters']['time_series']['filepath'].endswith("populated")]
+        if len(building_list) == 0:
+            raise Exception("No Modelica files found. The UO SDK simulations may not have been successful")
+
+        param_template['buildings']['custom'] = building_list
+        param_template['buildings']['default']['ets_model_parameters']['indirect']['nominal_flow_district'] = district_nominal_mfrt.round(3)
+
+        with open(sys_param_filename, 'w') as outfile:
```
macintoshpie, line 286: maybe we could call `validate` before we save the file to make sure it's valid? Though I'm not sure if it's possible or likely that we construct an invalid sys params

macintoshpie (commented):
Awesome ⚡ Just a few comments/suggestions

I tested the cli locally using the PR's suggested command and the output looks good to me.

geojson_modelica_translator/system_parameters/system_parameters.py
```diff
@@ -204,3 +207,81 @@ def validate(self):
             results.append(error.message)
 
         return results
+
+    def csv_to_sys_param(self, scenario_dir: Path, feature_file: Path, sys_param_filename: Path, overwrite=True) -> None:
```
nllong, line 211: 👍

geojson_modelica_translator/system_parameters/system_parameters.py
nllong: Can we make `self.data` be more explicit (e.g., `self.param_template_data`)?

geojson_modelica_translator/system_parameters/system_parameters.py
```diff
@@ -204,3 +207,81 @@ def validate(self):
...
+                            self.measure_list.append(Path(item) / "building_loads.csv")
+                        elif str(item).endswith('_export_modelica_loads'):
+                            self.measure_list.append(Path(item) / "modelica.mos")
+
+        # Parse the FeatureFile
+        building_ids = []
+        with open(feature_file) as json_file:
+            sdk_input = json.load(json_file)
+            for feature in sdk_input['features']:
+                if feature['properties']['type'] != 'Site Origin':
```
nllong, line 252: ideally this would use `jsonpath_ng` to find the fields.

geojson_modelica_translator/system_parameters/system_parameters.py
```diff
@@ -204,3 +207,81 @@ def validate(self):
...
+            feature_info = deepcopy(param_template['buildings']['custom'][0])
+            feature_info['geojson_id'] = building
+            building_list.append(feature_info)
+
+        # Grab the modelica file for the each Feature, and add it to the appropriate building dict
+        district_nominal_mfrt = 0
+        for building in building_list:
+            building_nominal_mfrt = 0
+            for measure_file_path in self.measure_list:
+                feature_name, measure_load_name = str(measure_file_path).split('/')[-3:-1]
```
nllong, line 267: can you document the `-3:-1` portion of this line...

management/uo_des.py
nllong: The template should be configurable outside of the SystemParameters class since the template can change if dealing with time series vs spawn vs others. However, I do think that Ted's point is salient--how does the user know what template to use?

nllong (changes requested):
Looking good! Few comments that should be addressed.

geojson_modelica_translator/system_parameters/system_parameters.py
vtnate: I'm working on loading the template in a different manner that won't re-use `self.data` to address this.

geojson_modelica_translator/system_parameters/system_parameters.py
vtnate: Abandoned this avenue. jsonpath_ng didn't work, for unknown reasons.

vtnate:
@nllong @macintoshpie I've updated this PR, can you review again?

management/uo_des.py
```diff
@@ -1,48 +1,56 @@
...
+    default='time_series',
     required=True
 )
-def sys_param(sys_param_filename, scenario_dir, sys_param_template, feature_file):
-    """GMT CLI."""
-    csv_to_sys_param = CSVToSysParam(
+def sys_param(model, sys_param_filename, scenario_dir, feature_file, overwrite):
+    """Create system parameters file using uo_sdk output"""
+    SystemParameters.csv_to_sys_param(
+        sys_param_template=model,
```
macintoshpie, line 51:
```suggestion
        model_type=model,
```

geojson_modelica_translator/system_parameters/system_parameters.py
```diff
@@ -204,3 +207,89 @@ def validate(self):
...
+                # Grab the relevant 2 components of the path: feature name and measure folder name, items -3 & -2 respectively
+                feature_name, measure_folder_name = str(measure_file_path).split('/')[-3:-1]
+                if feature_name != building['geojson_id']:
+                    continue
+                if (measure_file_path.suffix == '.mos'):
+                    building['load_model_parameters']['time_series']['filepath'] = str(measure_file_path.resolve())
+                if (measure_file_path.suffix == '.csv') and ('_export_time_series_modelica' in str(measure_folder_name)):
+                    mfrt_df = pd.read_csv(measure_file_path)
+                    building_nominal_mfrt = mfrt_df['massFlowRateHeating'].max()
+                    building['load_model_parameters']['time_series']['nominal_flow_building'] = float(building_nominal_mfrt)
```
macintoshpie, line 283: This doesn't appear to be in our current schema, should we add it?
Maybe we could add a test as well that runs the json schema validator on the generated result

management/uo_des.py
vtnate: Damn. You're right, of course.

geojson_modelica_translator/system_parameters/system_parameters.py
vtnate: ermuhgerd, yes, definitely.

nllong: has this been added?

nllong (approved):
changes look great!

One question that probably needs resolved.

geojson_modelica_translator/system_parameters/system_parameters.py
vtnate: > has this been added?

Not yet. Working on it now.

vtnate: This was resolved in the recent schema changes, I just needed to change a variable name slightly.

vtnate:
a44406c rearranges the sys-param file, bringing the ets parameters inside the custom building block.

geojson_modelica_translator/system_parameters/time_series_template.json
```diff
@@ -0,0 +1,35 @@
...
+      {
+        "geojson_id": "0",
+        "load_model": "time_series",
+        "load_model_parameters": {
+          "time_series": {
+            "filepath": "To be populated",
+            "delta_temp_air_cooling": 10
+          }
+        },
+        "ets_model_parameters": {
```
nllong, line 17: this looks good!

nllong (approved):
Thanks for updating the location of the ETS data.

nllong:
Hey @vtnate -- I think this is ready to merge. To make sure, we will address cleaning up the other sys param files in #284 as a separate PR?

vtnate:
> Hey @vtnate -- I think this is ready to merge. To make sure, we will address cleaning up the other sys param files in #284 as a separate PR?

Yes. #284 is low-risk, except for the one concern I raised in that PR.

nllong:
> > Hey @vtnate -- I think this is ready to merge. To make sure, we will address cleaning up the other sys param files in #284 as a separate PR?
> 
> Yes. #284 is low-risk, except for the one concern I raised in that PR.

cool cool. merging! w00t w00t. 🤘

PR #277, opened 2021-01-09
Fix simulation settings

AntoineGautier:
With the fix, the model `tests/model_connectors/output/district_cooling_system/Districts/DistrictEnergySystem.mo` now takes 15 s to simulate with JModelica.
Note that the simulation time could be lowered down to about 10 s by setting `ncp = 500` instead of `5000` currently. The latter value is only needed for year-long simulations.


nllong (approved):
This is great, thanks @AntoineGautier!

Issue #279, opened 2021-01-11
Replicate Sys Param Files From Time Series Template to All Others


laurenklun:
Ready but... scissors over paper:
Wait to merge until all tests are passing.

PR #284, opened 2021-01-11
Sys param rearranging

vtnate:
This is followup work to #273 
Moves the `ets_model_parameters` block of a sys-param file inside the building block, so there is an ets for each building.
`py.test`

However, `py.test tests/system_parameters/test_system_parameters.py::SystemParametersTest::test_get_param_with_none_building_id` fails with:

`AssertionError: {'indirect': {'heat_flow_nominal': 8000, [534 chars] 20}} != 'Not None'`

I suspect the failure is due to [this intentional behavior](https://github.com/urbanopt/geojson-modelica-translator/blob/sys-param-rearranging/geojson_modelica_translator/system_parameters/system_parameters.py#L188-L192). Potential ways of handling this:

1. Don't add the ets block to that portion of the test sys-param file (revert 2306fd3)
2. Change the test to expect different results
3. Change the actual code to not return defaults if missing a building_id

#279


nllong:
I like option 3 the most? What do you think?

vtnate:
> I like option 3 the most? What do you think?

I agree. What happens when we don't return defaults? Do we want just an information message and then return defaults? Message and quit?

nllong:
> > I like option 3 the most? What do you think?
> 
> I agree. What happens when we don't return defaults? Do we want just an information message and then return defaults? Message and quit?

Yes, just quit with a helpful message.

nllong (approved):
looks great! i updated the conflicts and this shoulfd be good now

PR #302, opened 2021-02-04
Cli incremental improvements

vtnate:

- Uses positional arguments instead of labeled options
    - Includes basic type-checking of user input
- Enables using short flags such as `-h` as well as `--help`


management/uo_des.py
```diff
@@ -36,50 +36,62 @@
...
-    "-f",
-    "--feature_file",
-    help="Path to urbanopt sdk FeatureFile, probably in the root UO SDK project directory.",
-    required=True
+@click.argument(
+    "feature_file",
+    type=click.Path(exists=True)
+)
+@click.argument(
+    "model-type",
```
macintoshpie, line 62: looks like it might not matter, but I see this is `model-type` vs `model_type`

macintoshpie (approved):
Looks good to me 👍  Nice improvements

management/uo_des.py
vtnate: Correct, Click adjusts all dashes in arguments to underscores.

Issue #296, opened 2021-01-29
Add assertions on test-generated .mat files


macintoshpie:
### Expected Behavior
GMT tests should assert that the values output by the simulated district model are valid.

### Actual Behavior
GMT tests only assert that the simulation was successfully run by JModelica (ie compiles and can finish without failing)

macintoshpie:
@Helarga @nllong We need to discuss which variables we want to "test". I know Nick mentioned total energy at one point, but I'm not sure where that would be derived.

Just for context, we have access to any (?) component/subcomponent in the district model. For example, `TimeSerLoa_e0ff8fd8.ports_aHeaWat[1].m_flow` and `TimeSerLoa_e0ff8fd8.ports_aChiWat[1].m_flow`
![image](https://user-images.githubusercontent.com/18518728/106323926-78898800-6235-11eb-8871-b91af846d670.png)
(image from model generated by test_district_heating_and_cooling_systems.py)

amyeallen1:
@macintoshpie Hi Ted, that's a good idea. Were you thinking of the check as being a sanity check that certain outputs are reasonable, or comparing to another set of specific values? I think the sanity checks would be useful, and, for the time series-based load models, ensuring that the boundary conditions imposed by the load are met. For sanity checks, we could check that values like mass flow rates and rates of heating delivered  (or heat rejected, if that's defined to be positive) are non-negative, and don't exceed their nominal values by more than a certain fraction, to give a range. 

In the case of the mass flow/temperature time series model, we could compare the supply temperature of the water delivered by the ETS to the supply temperature in the time series file. Per a previous discussion with Hagar, the supply temp delivered by the ETS is in the variable [name of ETS].sta_b2.T. In the ideal case, these two serieses of values are very consistent with each other. We could look at a metric like CV(RMSE) or something else to compare them. 

For the time series loads characterized with values in terms of overall thermal load (W), we could make an analogous comparison between the ETS.q_flow variable and the load time series, to ensure that the load is being met.  @Helarga @nllong  @macintoshpie

macintoshpie:
Thanks @amyeallen1 . I will open a PR adding these tests, probably starting with just the original TimeSeries load first. I wonder which tests we want to add this to? E.g. we have some that don't include plants, some that do, some that are heating or cooling indirect only, etc. Would we want these sanity checks added to all of of them, making assertions only about the parts included in the model? Or just to the full district system models?

[...]

PR #305, opened 2021-02-05
test: add model output assertions

macintoshpie:
GMT does not make assertions about model outputs after running the systems in JModelica

Adds validation of mass flow rate and Q flow to the district heating and cooling test for timeseries loads.

N/A

#296 

N/A


tests/model_connectors/test_district_heating_and_cooling_systems.py
```diff
@@ -111,3 +122,70 @@ def test_district_heating_and_cooling_systems(self):
...
+        cooling_indirect = cool_etses[0]
+        (_, heating_indirect_q_flow) = mat_results.values(f'{heating_indirect.id}.Q_flow')
+        (_, cooling_indirect_q_flow) = mat_results.values(f'{cooling_indirect.id}.Q_flow')
+        (_, load_q_heat_flow) = mat_results.values(f'{load.id}.QHea_flow')
+        (_, load_q_cool_flow) = mat_results.values(f'{load.id}.QCoo_flow')
+
+        cool_cvrmsd = cvrmsd(cooling_indirect_q_flow, load_q_cool_flow)
+        heat_cvrmsd = cvrmsd(heating_indirect_q_flow, load_q_heat_flow)
+
+        CVRMSD_MAX = 0.55
```
macintoshpie, line 183: Not sure what a good value for this is, we might need to test smaller values.

tests/model_connectors/test_district_heating_and_cooling_systems.py
```diff
@@ -111,3 +122,70 @@ def test_district_heating_and_cooling_systems(self):
...
+                return np.sqrt(
+                    np.sum(
+                        np.square(
+                            a - b
+                        )
+                    ) / n_samples
+                )
+
+            normalization_factor = max(a.max(), b.max()) - min(a.min(), b.min())
+            return rmsd(a, b) / normalization_factor
```
macintoshpie, line 170: @amyeallen1 Check my math? Also I used range between the two time series as the normalization factor, not sure what's actually appropriate to use here so looking for feedback.

tests/model_connectors/test_district_heating_and_cooling_systems.py
```diff
@@ -111,3 +122,70 @@ def test_district_heating_and_cooling_systems(self):
...
+
+        # check the mass flow rates of the first load are in the expected range
+        load = loads[0]
+        (_, heat_m_flow) = mat_results.values(f'{load.id}.ports_aHeaWat[1].m_flow')
+        (_, cool_m_flow) = mat_results.values(f'{load.id}.ports_aHeaWat[1].m_flow')
+        self.assertTrue((heat_m_flow >= 0).all(), 'Heating mass flow rate must be greater than or equal to zero')
+        self.assertTrue((cool_m_flow >= 0).all(), 'Cooling mass flow rate must be greater than or equal to zero')
+
+        # this tolerance determines how much we allow the actual mass flow rate to exceed the nominal value
+        M_FLOW_NOMINAL_TOLERANCE = 0.01
```
macintoshpie, line 141: Not sure what a good value for this is

macintoshpie (commented):
@nllong @amyeallen1 @Helarga 
Could you all take a look at this? So far I only added the test for `test_district_heating_and_cooling_system.py` which has two timeseries loads with heating and cooling indirect ETSes.

This is pretty verbose right now, will probably bundle it into helper functions when we add these tests to other files. @nllong I think we want to add the test for TimeSeriesMFT district system, but do we also want tests for models where we've stubbed out ETSes or Networks/Plants?

nllong:
> @nllong @amyeallen1 @Helarga
> Could you all take a look at this? So far I only added the test for `test_district_heating_and_cooling_system.py` which has two timeseries loads with heating and cooling indirect ETSes.
> 
> This is pretty verbose right now, will probably bundle it into helper functions when we add these tests to other files. @nllong I think we want to add the test for TimeSeriesMFT district system, but do we also want tests for models where we've stubbed out ETSes or Networks/Plants?

This looks great to me. I didn't double check your CVRMSE math (I'll let @amyeallen1) do that, but the overall approach and adding in the tolerance looks right. Probably set the tolerance a bit higher (e.g., 0.1). 

Yes on the helper functions as that will allow us to reuse it in the MFT. I think we should only test the full models for now (not the stubbed models).

tests/model_connectors/test_district_heating_and_cooling_systems.py
amyeallen1: Hi Ted, thanks for implementing this! I looked at the approach for calculating CV(RMSE) from ASHRAE's Guideline 14, which is often used for assessing the performance of energy models. They use the mean of the measured value for normalization, and (not that it makes much of a difference in the calculated value) they use (n-p) instead of n. They suggest a value of p=1 for "calibrated simulations." These component models aren't calibrated for a given set of time series data, but I thought that characterization fit the model better than the other types of models Guide 14 addresses (like change-point models). @macintoshpie @nllong 
image

tests/model_connectors/test_district_heating_and_cooling_systems.py
amyeallen1: @macintoshpie Hi Ted, thanks for doing this. As far as the threshold for CV(RMSE), I consulted ASHRAE Guideline 14, and for models to predict whole-building level electrical demand, they recommend an upper threshold of 30-40%, depending on the amount of baseline data available. (Since these component models aren't calibrated for the particular set of time series data used, I don't think that distinction is that relevant.) I couldn't find any specific guidance from G14 regarding models to predict other quantities (such as temperature in this case) as opposed to energy-related quantities, but based on the time step, I think demand is a better analogue for our case than energy consumption. Using a sample set of time series load data I generated from a prototype building model as an example and calculating CV(RMSE) from the results of the GMT-generated model per G14's approach, I got a CV(RMSE) of 35%. Maybe 40% would be an appropriate upper threshold? @nllong @Helarga 
image

amyeallen1 (commented):
Thanks, Ted, for making these modifications! Not that it's immensely consequential to the result, but as I interpret their definitions, G14 seems to use the mean of the measured values, and not the mean of the measured and simulated values, as the normalization factor. 
image

macintoshpie:
> G14 seems to use the mean of the measured values, and not the mean of the measured and simulated values, as the normalization factor

Ah ok thanks @amyeallen1 , I'll update it

amyeallen1:
@macintoshpie Thanks, Ted! In this case, the "measured" would be the data supplied through the time series file, and "simulated" would be the output from the model. Is that what you're getting at in terms of labeling?

macintoshpie:
@nllong python question for you
now that we're on poetry, its failing to install our own dependencies dependencies correctly. For buildingspy, Poetry is not installing scipy (and some other stuff), looks like because of the `--only-binary` option in their readme.txt? Any idea what's up with this? I don't really want to add scipy to our dependencies b/c we don't directly use it.

https://github.com/lbl-srg/BuildingsPy/blob/f9167f088825d07c4cfdd7dcd82475e5f60521c2/requirements.txt#L6

amyeallen1 (commented):
Thanks, Ted, for updating the normalization factor! Since there's not a multiplication by 100 (unless I'm missing it), I think we expect the CV(RMSE) values to generally be <1.  A CVRMSD_MAX of 10 seems high. If we are on the decimal scale, what about something like 0.4?

macintoshpie:
@amyeallen1 There might be something wrong with the math, but I'm getting the following values:
```python
cool_cvrmsd = -2.796775012423726
heat_cvrmsd = 6.5151145263289925
```
In fact, should we also be taking absolute value of both to make sure they're positive? The negative sign is coming from the fact that we're dividing RMSD by the mean of measured, which is negative for cooling q_flow.

Could you take another look at what I've got? Or could you suggest a function I could use inplace of my implementation? Happy to meet with you to chat about this if it'd make things easier.

amyeallen1:
@macintoshpie Thanks for thinking about this. That's a good point regarding absolute value--yes, since we expect one set of the loads (cooling) to be defined as being negative, since they represent a removal of heat, I think we should take absolute value of both the measured and modeled serieses. 

Sorry for not having run the tests before (I was doing a hand calc of CV(RMSE))--when running the tests from this branch, I don't see any output regarding the assertion about CV(RMSD) printed to the command line. Is it supposed to be, or where should I find those results? @macintoshpie

amyeallen1:
Hi Ted, sorry for not thinking of this before when you mentioned the absolute value issue. The difference in sign between the Q flow from the ETS and the load values are the reason for the high magnitude of CV(RMSD). By way of explanation, the load values supplied by the user are negative for cooling (for rejection of heat) and positive for heating, and the load outputs from the ETS follow the opposite convention. I suggest we take the absolute value of the expected negative values (the measured cooling load and modeled heating load) (and not all the values, so we don't camouflage any unexpected signs). Then, the threshold for CVRMSD could be 0.4 or 0.5. Thanks! @macintoshpie

macintoshpie:
@amyeallen1 I took absolute value of all of the q flows (640ca08) and they definitely got smaller, but not as small as you mentinoed
```
(Pdb) cool_cvrmsd
0.10656371834521426
(Pdb) heat_cvrmsd
2.7427656245810583
```

I just set the CVRMSD_MAX to 3 for now, do you think this is ok?

nllong:
based on Guideline 14: `The computer model shall have an NMBE of 5% and a CV(RMSE) of15% relative to monthly calibration data. If hourly calibration data are used, these requirements shall be 10% and 30%, respectively`, and assuming that you are in percent, then yes. Otherwise, if you are absolute, then the heating is off by over 274%, which is really bad.

amyeallen1:
Thanks, Ted, for implementing the changes. Right, Nick, that is 274% for the heating comparison. 

The CV(RMSE) calculation looks correct. I ran the test with the load profiles generated directly from the prototype office building for 5B, and confirmed the CV(RMSE) for heating with a hand calculation, which was around 30%. Perhaps for now, Ted could make the CV(RMSE) threshold a warning instead of an error, based on the 30% value, or some margin above that, and I could investigate more why the model isn't able to track the heating load well in this case? 
@nllong @macintoshpie

macintoshpie:
I went ahead and made it a warning. Created this issue here for tracking the heating q flow fix and assigned to you Amy: #311 

@nllong could you review this and if things look good let's merge?

amyeallen1:
@macintoshpie Thanks, Ted!

pyproject.toml
```diff
@@ -23,7 +23,7 @@ packages = [
 [tool.poetry.dependencies]
 python = "^3.7"
 click = "7.1.2"
-BuildingsPy = "2.0.0"
+BuildingsPy = "2.1.0"
```
nllong, line 26: noice

nllong (approved):

Issue #264, opened 2020-12-21
Dymola Component Layout


laurenklun:
1/4: postpone until we have more time to look at this. 

look at, understand problems, tell team and then figure out what to do. 

check out linkages.js world. Idea of presenting modelica files in a visually readable layout.

macintoshpie:
See #268 and #333

PR #315, opened 2021-02-24
Feat/visual graph

macintoshpie:
Modelica allows models to be rendered as diagrams (icons and connecting lines). The GMT does not do anything with this currently, so components are placed messily at the DHC level.

Adds ability to templatize icon positions and lines. Generate placement of icons using `{{ diagram.transformation.<name>.<type> }}`. Generate line connections with `{{ diagram.line.<name a>.<port a>.<name b>.<port b> }}` 

This PR adds this templating for all models. This current implementation _only_ places icons, and it does so in an arbitrary way by placing one in every other "grid cell".

Note that the rendering is NOT how we want the diagram to look. This is just a first step in proving how it can be done. 
Note that this rendering does NOT handle connections (connections are parsed, but we are not organizing, templating them)
Another task will be to actually organize the components how we like, then route the connecting lines.

Another limitation of the current implementation is that it assumes all icons are the same dimensions, which they are not (see the network icons). This will be addressed later as well.

Run `test_district_heating_and_cooling_systems.py` and load the generated model in Dymola.

#264 

Screen Shot 2021-03-12 at 3 59 14 PM
Screen Shot 2021-03-12 at 3 58 24 PM
Screen Shot 2021-03-12 at 3 57 37 PM


geojson_modelica_translator/model_connectors/couplings/coupling.py
```diff
@@ -171,3 +171,11 @@ def render_templates(self, template_params):
...
             'connect_statements_template_path': connect_template_path,
         }
+
+    @property
+    def component_definitions_template_path(self):
+        return self._template_env.get_template(self._template_component_definitions).filename
+
+    @property
+    def connect_statements_template_path(self):
+        return self._template_env.get_template(self._template_connect_statements).filename
```
macintoshpie, line 181: Added these to make it convenient to find the template files for a coupling so we could parse out the places where we want to insert diagram things

geojson_modelica_translator/model_connectors/couplings/diagram.py
```diff
@@ -0,0 +1,227 @@
+"""
```
macintoshpie, line 1: will probably move this and related files into the district directory

geojson_modelica_translator/model_connectors/model_base.py
```diff
@@ -54,9 +54,14 @@ def __init__(self, system_parameters, template_dir):
 
         # initialize the templating framework (Jinja2)
         self.template_dir = template_dir
-        self.template_env = Environment(loader=FileSystemLoader(searchpath=self.template_dir))
+        self.template_env = Environment(
+            loader=FileSystemLoader(searchpath=self.template_dir),
+            undefined=StrictUndefined
```
macintoshpie, line 59: ⚠️ Important! I added `StrictUndefined` here which will make templating fail if it's missing a variable in the template context.

geojson_modelica_translator/model_connectors/couplings/diagram.py
```diff
@@ -0,0 +1,227 @@
...
+        def translate_x(pos):
+            # translate from origin at upper left of grid to center of grid
+            return pos - (self.grid_size / 2)
+
+        def translate_y(pos):
+            return (self.grid_size / 2) - pos
+
+        transformations = defaultdict(dict)
+        lines = defaultdict(dict)
+        transformation_template = Template('transformation(extent={{$x1,$y1},{$x2,$y2}})')
```
nllong, line 87: I was unaware that python's string had this method, very cool!

geojson_modelica_translator/model_connectors/couplings/utils.py
```diff
@@ -0,0 +1,51 @@
+import re
+from collections import namedtuple
+
+DiagramTransformation = namedtuple('DiagramTransformation', ['model_name', 'model_type'])
+DiagramLine = namedtuple('DiagramLine', ['a_name', 'a_port', 'b_name', 'b_port'])
+
+JINJA_EXPRESSION_REGEX = re.compile(r'{{\s*(.*?)\s*}}')
```
nllong, line 7: fancy

geojson_modelica_translator/model_connectors/couplings/diagram.py
nllong: Does this mean that the diagram would live with the district system only? If we want a different diagram for a subcomponent (e.g., the TEASER file), then we would add a diagram.py file to that path? I'm just trying to understand how this will manifest in the future.

nllong (commented):
I think this is a great start Ted! 

* I like how you are able to layout the components in the grid
* The overall approach makes sense
* This is definitely the first step and we need to decide on how to group similar groups of components based on our other discussions. For example--sources should be on the left, sinks on the right.
* Obviously making the connection lines not overlap needs to be investigated. But I think that is a much lower priority.

I would like to get other feedback on this approach, specifically from Dymola users.

geojson_modelica_translator/model_connectors/couplings/diagram.py
macintoshpie: That's a good question. I definitely had it in my head that we would only be modifying the diagram layer at the district level. Thanks for bringing this up, I might not move diagram.py

macintoshpie:
> I would like to get other feedback on this approach, specifically from Dymola users.

Anything I can help with on this? Do you mean checking with them for our placement of icons?

nllong:
> > I would like to get other feedback on this approach, specifically from Dymola users.
> 
> Anything I can help with on this? Do you mean checking with them for our placement of icons?

I think there are two things additional types of reviews that I would like. 1) does this make sense from the Python side. Perhaps we loop in @kbenne to present this plan to and get his reaction. He has experience laying out HVAC systems in OpenStudio. And 2) looping in @AntoineGautier on the Dymola side and see his reaction on how this is structured.

macintoshpie:
@nllong I wrapped up adding the templating to the rest of the district-level components. I had to fix a bug I found with templating in lines as well.

Per our discussion, I've created #333 to organize the icons rather than arbitrarily placing them. I think this is ready to merge if things look good on your end

nllong (approved):
great work here! Thanks Ted!

Issue #194, opened 2020-10-05
[4/15] Documentation Site Content Buildout


laurenklun:
release current version with poetry
- teaser issues... need to update modelica builder

nllong:
Let us discuss on the next call this document: https://docs.google.com/document/d/1IwoxEY8ZcEuhZZF68RujD3faAWLKY57-/edit?usp=drive_web&ouid=105632665010758441410&rtpof=true

PR #318, opened 2021-03-04
Use latest release of TEASER

nllong:

It appears that poetry does not allow pushing to PyPI if a dependency is a GitHub repository. Even if there is a way around this, it was time to upgrade to the latest version of TEASER. However, the latest version still does not include the changes around the ROM variable names nor the internal gain time series files. 

* Upgrade to newest release of Modelica-builder (version 0.1.1) which adds the `rename_component_argument`
* Fix the ROM models and the internal gain files
* Point to 0.7.5 TEASER

* py.test should cover everything

This work is only part of the work needed to release a new version and to provide better documentation (#194) 

N/A

### DIFF
--- a/geojson_modelica_translator/model_connectors/load_connectors/teaser.py
+++ b/geojson_modelica_translator/model_connectors/load_connectors/teaser.py
@@ -344,8 +344,8 @@ class Teaser(LoadBase):
                 thermal_zone_name = 'thermalZoneFourElements'
 
             if thermal_zone_name is not None and thermal_zone_type is not None:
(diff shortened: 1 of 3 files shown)


macintoshpie (approved):
Looks great Nick! Pulled and ran locally and it updated Floor.mo as expected.

I _think_ I also had to remove the venv for it to use the "new" teaser package at `poetry env list --full-path`

nllong:
Yeah, I think poetry is a bit too aggressive in how it holds onto things.

PR #330, opened 2021-03-09
Cli expansion

vtnate:
Expands cli to build and run the model
adds 2 new cli commands
`poetry run uo_des -h` to see available commands. Try them all out.
- Ensure the `run-model` command works as expected
    - broken as of March 8, 2021. Internal gmt modelica variable names don't get found


nllong:
Hey @vtnate  -- I've just committed what I think will fix it. It had to do with the paths for sure--we have to make sure that we pass the entire modelica package to the run_in_docker command. That is, we pass the entire directory where the `package.mo` file lives. Also, the ZIP file that you sent did not have the `package.mo` more the `package.order` file. If those files do not exist, then the simulation also fails (with a different message than what you have been seeing).

Lastly, I removed the aggressive moving of all files after the simulation finished. It should only move stdout.log and {project_name}_* files now.

nllong (approved):

vtnate:
Beautiful @nllong! I made a few more cleanupy changes, the biggest of which is having users specify the scenario-file instead of the scenario-dir from the sdk. That's consistent with how the sdk commands run, so this should be less confusing this way. I also changed the name of the commands in the main help screen, so the alphabetical ordering is also the correct order to run them.

PR #339, opened 2021-03-16
fix path to modelica project dir

vtnate:
Apparently I didn't test enough situations for the cli earlier, and missed this path requirement
Makes path formulations for the modelica project dir explicit, enabling arbitrary location of project dir
Using the `create-model` cli command, give the `project_name` a path to somewhere that isn't pwd


nllong (approved):
thanks for fixing this!

PR #354, opened 2021-03-25
Lb bugfixes

vtnate:
LadyBug found several bugs when they ran the cli on Windows machines
- Changes how paths are handled
- Adds `overwrite` flag to `create-model` command
- Change dtype from numpy to float, so Windows will be happy with json.dump
- Adds test for overwrite flag
Try running the cli commands on Windows and non-Windows machines
Do my changes to coupling.py and model_base.py actually work on Windows?


management/uo_des.py
```diff
@@ -155,7 +156,14 @@ def build_sys_param(model_type, sys_param_filename, scenario_file, feature_file,
     "model_type",
     default='time_series',
 )
-def create_model(model_type, sys_param_file, geojson_feature_file, project_name):
+@click.option(
+    '-o',
+    '--overwrite',
+    is_flag=True,
```
chriswmackey, line 162: This is just a style suggestion that you don't have to implement but you can automatically make something a boolean flag by formatting the name of the input like you see here:
https://github.com/ladybug-tools/honeybee-energy/blob/master/honeybee_energy/cli/simulate.py#L46-L50

I tend to like it since having a flag for both of the two options usually feels a bit more explicit. But, again, totally up to you.

management/uo_des.py
```diff
@@ -37,6 +37,7 @@
 """
 
 from pathlib import Path
+from shutil import rmtree
```
chriswmackey, line 40: This should usually work but I have had a little more success with the following rather than using `shutil`:
```
import distutils
distutils.dir_util.remove_tree(path_to_folder)
```

I think it might just have better ways to get around the deletion of folders that are open in Windows Explorer or Mac Finder.

geojson_modelica_translator/system_parameters/system_parameters.py
```diff
@@ -112,14 +111,14 @@ def resolve_paths(self):
...
+        filepath = Path(self.filename).parent.resolve()
 
         for pe in self.PATH_ELEMENTS:
             matches = parse(pe["json_path"]).find(self.data)
             for index, match in enumerate(matches):
                 # print(f"Index {index} to update match {match.path} | {match.value} | {match.context}")
-                new_path = os.path.join(filepath, match.value)
-                parse(str(match.full_path)).update(self.data, new_path)
+                new_path = Path(filepath) / match.value
+                parse(str(match.full_path)).update(self.data, str(new_path))
```
chriswmackey, line 121: I'm just pasting @mostaphaRoudsari 's suggestion here:

```
from pathlib import Path
path = Path(r'c:\ladybug')
print(path.as_posix())  # prints c:/ladybug
print(str(path))  # prints c:\ladybug
```

chriswmackey (commented):
Generally looking good @vtnate . There's only one comment that I think is a bug and the rest are just suggestions.

geojson_modelica_translator/system_parameters/system_parameters.py
```diff
@@ -294,7 +294,7 @@ def csv_to_sys_param(cls, model_type: str, scenario_dir: Path, feature_file: Pat
             raise Exception("No Modelica files found. The UO SDK simulations may not have been successful")
 
         for building in building_list:
-            building['ets_model_parameters']['indirect']['nominal_mass_flow_district'] = district_nominal_mfrt.round(3)
+            building['ets_model_parameters']['indirect']['nominal_mass_flow_district'] = float(district_nominal_mfrt.round(3))
```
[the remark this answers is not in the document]
vtnate: Fascinating. Line 287 doesn't work to cast to float, so I changed line 288 instead which should have the same effect.

management/uo_des.py
vtnate: Thanks for showing me that @chriswmackey. I will keep it as is, as I like the idea of including the flag for true and not for false. Good to know this method also exists!

management/uo_des.py
vtnate: Very interesting @chriswmackey. I'm hesitant to use distutils because the first 2 paragraphs of the docs make it sound like this isn't the right use for that package: https://docs.python.org/3/library/distutils.html.

Do you think the same effect could be had by passing `ignore_errors=True` like this:
`shutil.rmtree(path, ignore_errors=True)`?

geojson_modelica_translator/system_parameters/system_parameters.py
```diff
@@ -112,14 +111,14 @@ def resolve_paths(self):
...
+        filepath = Path(self.filename).parent.resolve()
 
         for pe in self.PATH_ELEMENTS:
             matches = parse(pe["json_path"]).find(self.data)
             for index, match in enumerate(matches):
                 # print(f"Index {index} to update match {match.path} | {match.value} | {match.context}")
-                new_path = os.path.join(filepath, match.value)
-                parse(str(match.full_path)).update(self.data, new_path)
+                new_path = Path(filepath) / match.value
+                parse(str(match.full_path)).update(self.data, str(new_path.as_posix()))
```
mostaphaRoudsari, line 121: `as_posix` returns a string.

```suggestion
                parse(str(match.full_path)).update(self.data, new_path.as_posix())
```

vtnate: So it does. My earlier test config must have been different in some other way. Thanks Mostapha.

vtnate:
@macintoshpie Can we set up Github Actions to test with Windows as well as Linux? Is that a thing?

management/uo_des.py
chriswmackey: I understand, @vtnate . My suggestion came from anecdotal experience rather than an official source so we definitely don't need to implement it.  I just put it here in case we eventually realize that `shutil` isn't working as intended. So we can try it  with `shutil.rmtree(path, ignore_errors=True)` for now and, if we see it becoming an issue, we can always change it later.

chriswmackey:
Looks good to me @vtnate !

vtnate:
@macintoshpie Can you help me fix the test I wrote? I compared it to https://github.com/urbanopt/geojson-modelica-translator/blob/develop/tests/system_parameters/test_system_parameters.py#L248-L263 and can't figure out why my new test isn't working locally.

vtnate:
@mostaphaRoudsari Will you please review this again? Make sure all the cli commands work as expected on your Windows machine. If so, please formally approve the PR and I'll merge it in!

geojson_modelica_translator/model_connectors/couplings/coupling.py
```diff
@@ -168,9 +168,9 @@ def _get_model_id(model):
 
         # get the template file path relative to the package
         template_filename = template.filename
-        _, template_filename = template_filename.rsplit('geojson_modelica_translator/', 1)
+        _, template_filename = template_filename.rsplit('geojson_modelica_translator', 1)
 
-        return template.render(updated_template_params), template_filename
+        return template.render(updated_template_params), template_filename[1:]
```
mostaphaRoudsari, line 173: Is the [1:] for trying to remove the extra `/` at the end of the path? In that case it might not work with a path with `\\`. Maybe you can use pathlib.Path here.

```python
        template_filename = pathlib.Path(template.filename).as_posix()
        _, template_filename = template_filename.rsplit('geojson_modelica_translator/', 1)
        return template.render(updated_template_params), template_filename
```

geojson_modelica_translator/model_connectors/model_base.py
```diff
@@ -167,9 +167,9 @@ def render_instance(self, template_params):
 
         # get template path relative to the package
         template_filename = template.filename
-        _, template_filename = template_filename.rsplit('geojson_modelica_translator/', 1)
+        _, template_filename = template_filename.rsplit('geojson_modelica_translator', 1)
 
-        return template.render(template_params), template_filename
+        return template.render(template_params), template_filename[1:]
```
mostaphaRoudsari, line 172: same comment here.

geojson_modelica_translator/system_parameters/system_parameters.py
```diff
@@ -71,13 +70,13 @@ def __init__(self, filename=None):
         :param filename: string, (optional) path to file to load
         """
         # load the schema for validation
-        with open(os.path.join(os.path.dirname(__file__), "schema.json"), "r") as f:
+        with open((Path(__file__).parent / "schema.json"), "r") as f:
             self.schema = json.load(f)
```
mostaphaRoudsari, line 74: You can use Path `read_text` method to read the file. Both code will work but this is technically using the OOP goodies of Path.

```suggestion
        schema = Path(__file__).parent / "schema.json"
        self.schema = json.loads(schema.read_text())
```

mostaphaRoudsari (commented):
Hi @vtnate, There is one case that might become a problem. Otherwise I think it should work. @chriswmackey is the right person to test the command line.

vtnate:
@chriswmackey This PR should resolve most of our problems by now. Can you pull the newest and try this out on your machine?

However, running simulations in Windows with our CI is still a thorn in our sides. We've found a problem and considered a few solutions in #360 (some easier/faster to implement than others).

vtnate:
@macintoshpie These last 3 `test (3.7...` checks don't seem to be finishing/running. Is that related to the skipped simulations on Windows?

macintoshpie:
@vtnate It's because we used to run only those tests (OS wasn't part of the test matrix), and in our repo settings we have it as required before merge.

@nllong can you remove `test (3.7 docs)` `test (3.7 precommit)` and `test (3.7 python)` from the required checks in the repo settings?

nllong:
> @vtnate It's because we used to run only those tests (OS wasn't part of the test matrix), and in our repo settings we have it as required before merge.
> 
> @nllong can you remove `test (3.7 docs)` `test (3.7 precommit)` and `test (3.7 python)` from the required checks in the repo settings?

dunzo

chriswmackey:
Hey @vtnate ,

I verified that the current fixes allow me to get all of the way up though the creation of the Modelica input files on Windows. So I think this is good to merge in my opinion.  I will deal with the issues of getting docker to run on Windows after this is merged.

FYI, the docs here seem to be out of date:
https://github.com/urbanopt/geojson-modelica-translator#getting-started

It looks like the command should be 
```
uo_des create-model sys_param.json example_project.json
```

vtnate:
Good catch, thanks @chriswmackey. I'll update the docs. Thanks for your review! We'll figure out what to do about simulations.

nllong (approved):

Issue #357, opened 2021-03-29
Updating the Dist H&C Test


laurenklun:
updating the time series that are compared for the RMSE.

See #311

PR #358, opened 2021-03-29
test: update district H+C q flow assertions

macintoshpie:
Uses Load's QReq[Coo || Hea]_flow instead of ETS's Q_flow for thermal load assertion

na
#357


macintoshpie:
@amyeallen1 @nllong CVRMSE is still bigger than we'd like for the heating CVRMSE, It reported a value of `CVRMSD=0.7320269339337164` when run locally (we want `< 0.3`)

nllong:
> @amyeallen1 @nllong CVRMSE is still bigger than we'd like for the heating CVRMSE, It reported a value of `CVRMSD=0.7320269339337164` when run locally (we want `< 0.3`)

But we are moving in the right direction! I think before we were in the 250 (not percent) range.

amyeallen1:
Thanks, Ted, for making the revisions! @macintoshpie

PR #389, opened 2021-05-24
Identify the unused schema fields

### LINKED ISSUE #347: Update System Design Parameter Schema with Flags - Used/Not Used
Let CU know when this is done and how to access.

nllong:
The buildings and system parameters schema have many fields which are unused in the GMT. 

This PR added in an `[unused]` identifier to the fields that aren't currently read into the GMT templating framework.

* `cd docs`
* `make html`
*  Inspect the webpage
 
Resolves #347 

image

### DIFF
--- a/docs/_static/custom_sphinx_styles.css
+++ b/docs/_static/custom_sphinx_styles.css
@@ -27,6 +27,9 @@ div.sphinxsidebar ul, div.sphinxsidebar ul ul {
     list-style-type: none;
 }
 
+div.body {
+    max-width: 1600px;
+}
 div.footer, div.footer a {
     color: #2c3e50;
 }
--- a/geojson_modelica_translator/system_parameters/schema.json
+++ b/geojson_modelica_translator/system_parameters/schema.json
@@ -4,7 +4,7 @@
   "definitions": {
     "system_design_parameter_def": {
       "title": "System Design Parameter Schema",
-      "description": "Definition of System Design Parameters. (Version 0.1)",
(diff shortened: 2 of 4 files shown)


geojson_modelica_translator/system_parameters/schema.json
```diff
@@ -693,41 +693,41 @@
       ]
     },
     "central_heating_plant_parameters": {
```
nllong, line 695: It appears that we aren't actually setting any of the central heating plant sys parameters. Is this getting resolved with the CHP work @vtnate ?

docs/_static/custom_sphinx_styles.css
```diff
@@ -27,6 +27,9 @@ div.sphinxsidebar ul, div.sphinxsidebar ul ul {
     list-style-type: none;
 }
 
+div.body {
+    max-width: 1600px;
```
nllong, line 31: the width was causing many of the columns to overlap. There are still a few that are overlapping--this can be resolved by patching the sphinx jsonschema package, but I didn't want to go down that route.

macintoshpie (approved):
Looks good, 👍 

What is the `update_schemas` command used for?

nllong:
> Looks good, 👍
> 
> What is the `update_schemas` command used for?

it pulls in the schemas (not system parameter schema) from the broader URBANopt project.

geojson_modelica_translator/system_parameters/schema.json
vtnate: Correct, we have never templatized heating plants before, and I'm including that work in the CHP PR.

PR #394, opened 2021-06-07
fix and update poetry dependencies

### LINKED ISSUE #393: Issue with poetry build-system section when including as a dependency
When including the GMT as a dependency of another project, the `poetry install` command on the other project results in an error that this project's `pyproject.toml` file is incorrect.

```
  Command ['/Users/nlong/.pyenv/versions/3.7.9/envs/3.7.9-gmt-ex/bin/pip', 'install', '--no-deps', '-U', '/Users/nlong/working/urban-analysis/geojson-modelica-translator'] errored with the following return code 1, and output:
  Processing /Users/nlong/working/urban-analysis/geojson-modelica-translator
    DEPRECATION: A future pip version will change local packages to be built in-place without first copying to a temporary directory. We recommend you use --use-feature=in-tree-build to test your packages with this new behavior before it becomes the default.
     pip 21.3 will remove support for this functionality. You can find discussion regarding this at https://github.com/pypa/pip/issues/7555.
  ERROR: file:///Users/nlong/working/urban-analysis/geojson-modelica-translator has a pyproject.toml file that does not comply with PEP 518: 'build-system.requires' contains an invalid requirement: 'poetry = 1.1.6'


  at ~/.pyenv/versions/3.7.9/envs/3.7.9-gmt-ex/lib/python3.7/site-packages/poetry/utils/env.py:1075 in _run
      1071│                 output = subprocess.check_output(
      1072│                     cmd, stderr=subprocess.STDOUT, **kwargs
      1073│                 )
      1074│         except CalledProcessError as e:
    → 1075│             raise EnvCommandError(e, input=input_)
      1076│
      1077│         return decode(output)
      1078│
      1079│     def execute(self, bin, *args, **kwargs):
```

nllong:
We were using an older pyproject.toml section for `[build-system]`. This was causing an issue when requiring the package from another project.

New format is the following.
```
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"
```

* make sure everything runs (`poetry update`, `poetry run py.test`)

Fixes #393

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,6 +53,7 @@ sphinx-jsonschema = "1.16.7"
 toml = "0.10.2"
 tox = "3.20.0"
 
+[build-system]
 requires = ["poetry-core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,10 +53,8 @@ sphinx-jsonschema = "1.16.7"
 toml = "0.10.2"
 tox = "3.20.0"
 
-
-[build-system]
-requires = ["poetry = 1.1.4"]
-build-backend = "poetry.masonry.api"
+requires = ["poetry-core>=1.0.0"]
+build-backend = "poetry.core.masonry.api"
 
 [tool.poetry.scripts]
 uo_des = 'management.uo_des:cli'
--- a/poetry.lock
+++ b/poetry.lock
@@ -32,17 +32,17 @@ python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
 
 [[package]]
 name = "attrs"
-version = "20.3.0"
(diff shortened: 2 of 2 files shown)


nllong:
welp, this didn't solve it @macintoshpie. I thought it worked because it fixed the installation from the GMT-examples side, but then broke just installing the GMT by itself. ¯\_(ツ)_/¯

macintoshpie:
@nllong looks like there was a small typo in the pyproject.toml, but should be fixed now and works with the examples repo. If this looks good to you then I think we can merge this

pyproject.toml
```diff
@@ -53,6 +53,7 @@ sphinx-jsonschema = "1.16.7"
 toml = "0.10.2"
 tox = "3.20.0"
 
+[build-system]
```
nllong, line 56: 😲 😔

macintoshpie (approved):
👍

PR #396, opened 2021-06-10
Updates to test file to generate a mixed-loads model incl. the mass flow time series 

amyeallen1:
Hi, I'm opening this PR for feedback on an adjustment to one of the tests--not for this to be merged. 

I'd like to use the mixed_loads_test file as a starting point to generate a model with Spawn, Teaser, and mass flow time series load models as a starting point. The test will work (in terms of generating the model) with my Spawn and Teaser inputs, but I've run into a hiccup in modifying the test to use a mass flow time series model, instead of the W time series. (The hiccup also occurs using the MFT test data, so I think there's a problem with my approach beyond the input data.) I'd appreciate any light you can shed on this, Ted and Nate! 

I'm using the input files in the mixed_loads_district folder.


nllong (approved):
the model still has some issues, but merging since the MBL 9 is going to be a huge lift regardless.

PR #411, opened 2021-07-29
Alert user that Modelica doesn't support spaces in project name

### LINKED ISSUE #375: Broken file path if spaces
Check if this got fixed. 
if path has a space in the name our shell script breaks

vtnate:
If a user has a space in their project name or directory tree, Modelica chokes and dies.
Adds graceful error message to user requiring them to remove spaces from name and path.
Project name gets tested for a space in automated testing now. `poetry run py.test tests/management/test_uo_des.py`
To test the tree, change something in your dir tree to include a space and try running `uo_des run-model`
Resolves #375

### DIFF
--- a/geojson_modelica_translator/model_connectors/load_connectors/load_base.py
+++ b/geojson_modelica_translator/model_connectors/load_connectors/load_base.py
@@ -176,7 +176,7 @@ class LoadBase(ModelBase):
                         except KeyError:
                             floor_height = 3  # Default height in meters from sdk
                             print(
-                                f"\nNo floor_height found in geojson feature file for building {self.building_id}. "
+                                f"No floor_height found in geojson feature file for building {self.building_id}. "
                                 f"Using default value of {floor_height}.")
 
                         # UO SDK defaults to current year, however TEASER only supports up to Year 2015
--- a/management/uo_des.py
+++ b/management/uo_des.py
@@ -166,7 +166,8 @@ def create_model(sys_param_file, geojson_feature_file, project_path, overwrite):
             raise SystemExit(f"Output dir '{project_path}' already exists and overwrite flag is not given")
     if len(project_path.name.split()) > 1:  # Modelica can't handle spaces in project name
         raise SystemExit(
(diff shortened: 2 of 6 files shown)


macintoshpie (approved):
Nice! I wonder if we should also add this check inside of modelica_runner.py? Someone might not be using the CLI to run projects. Also maybe we should add these checks to [district.py](https://github.com/urbanopt/geojson-modelica-translator/blob/78962913d93086a492b3a1d16d1f944ba8fe5ebd/geojson_modelica_translator/model_connectors/districts/district.py#L70) for the same reason? What do you think?

management/uo_des.py
```diff
@@ -118,7 +118,7 @@ def build_sys_param(model_type, sys_param_filename, scenario_file, feature_file,
     if Path(sys_param_filename).exists():
         print(f"\nSystem parameters file {sys_param_filename} successfully created.")
     else:
-        raise Exception(f"{sys_param_filename} failed. Please check your inputs and try again.")
+        raise SystemExit(f"{sys_param_filename} failed. Please check your inputs and try again.")
```
vtnate, line 121: FYI: If we raise an `Exception` then a traceback is included in the output to the user. By raising a `SystemExit` we only output the custom error message to the user. Much cleaner this way.

geojson_modelica_translator/model_connectors/load_connectors/load_base.py
```diff
@@ -176,7 +176,7 @@ def add_building(self, urbanopt_building, mapper=None):
                         except KeyError:
                             floor_height = 3  # Default height in meters from sdk
                             print(
-                                f"\nNo floor_height found in geojson feature file for building {self.building_id}. "
+                                f"No floor_height found in geojson feature file for building {self.building_id}. "
```
vtnate, line 179: I've made so many adjustments to this formatting. I think this way is the least ugly while still remaining informative. Perhaps we don't need to tell the user this at all?

macintoshpie (approved):
Awesome 👍 ⚡

Issue #404, opened 2021-07-12
CHP Milestone Delivery


laurenklun:
Package for delivery to Bob Gemmer

1. be clear that this is residential scale
2. possibly line someone up to present on CHP and modelica

vtnate:
PPT has been completed and needs to be reviewed. I will contact @nllong outside of Github for this.

PR #416, opened 2021-08-06
bugfix from Mingzhe for chp

vtnate:
CHP wasn't making any electricity
- Disconnects CHP electricity from port A that never should have been
- Shows output from CHP electricity going to port B more explicitly (changed annotation)
- Sets a value for `preTem` that mistakenly was left out
Run the `test_chp_system.py` with pytest, run the generated model from the test in Dymola, inspect the `eleFol.PEleNet` output variable. Should look like the bottom screenshot in this comment.
#404 
District with CHP as central plant:
chp_district
Components of the CHP plant:
chp_plant
Electrical output from the CHP plant:
elec_output_from_chp_plant


macintoshpie (approved):
I can't comment on correctness but this seems good to me 👍 😎

Issue #403, opened 2021-07-06
Error when trying to use laboratory building type for TEASER model 


amyeallen1:
I tried to generate a TEASER model based on the "laboratory" building type, and had modified the TEASER template to do so, by extending the vector with zone-level nominal loads, to match the # of loads in the laboratory building, as opposed to the office building. After doing that, I got the following error message from the simulation. I couldn't track down the Internal Loads file that's being referenced, but this seems like the error may have something to do with an inconsistent table length. 

"ModelicaMessage category="warning"><value name="msg">"... loading ""Internals"" from ""/tmp/JModelica.org/jm_tmpH9EJii/resources/4/InternalGains_Storage.txt""&#10;"</value></ModelicaMessage>
FMIL: module = Model, log level = 3: [WARNING][FMU status:Warning] <ModelicaError category="warning"><value name="msg">"End-of-file reached when reading numeric data of matrix ""Internals(8760,22)"" from file ""/tmp/JModelica.org/jm_tmpH9EJii/resources/4/InternalGains_Storage.txt""&#10;"</value></ModelicaError>"

Issue #399, opened 2021-07-06
Suggested modifications to increase flexibility of TEASER template 


amyeallen1:
The current TEASER template [hard-codes](https://github.com/urbanopt/geojson-modelica-translator/blob/01c893c0d899c39a213d209c09ab76970802b0a0/geojson_modelica_translator/model_connectors/load_connectors/templates/TeaserBuilding.mot#L57
) the zone nominal loads, which are based on the number of zones in the office building model. This requires a modification by the user to represent another building type, such as a laboratory. Additionally, this template structure has the effect of limiting the user to representing a single building type through TEASER in a given project. 

It would be ideal if a) the zone nominal loads could be populated in a programmatic way, and b) the number of zones could also be set programmatically based on the building type, and c) if this could somehow be flexible to multiple bldg types in a given model.

PR #417, opened 2021-08-09
Modification to support different TEASER building types

amyeallen1:
This PR sets nominal zone loads for the TEASER model based on a vector, making it extensible to different building types. If the commonly occurring zone type with a significantly higher load density (the IT room, or "ICT" zone) is present, a higher nominal cooling load is set for it. This was previously hard-coded. 

I converted the nominal heating and cooling loads to strings in the Teaser.py file in order to implement this approach with the correct formatting in the .mot file. These nominal load values aren't used anywhere else that I could find, but if there is a way to do this string conversion in the template file itself, that would be more elegant. 

This has been tested specifically on the "Education", "Office," and "Laboratory" building types, which are ones that are likely to be used. 

This PR addresses the issues in tickets #399 and #403. Nick or Nate, would you like to review? @nllong @vtnate Thanks!

https://github.com/urbanopt/geojson-modelica-translator/issues/399

https://github.com/urbanopt/geojson-modelica-translator/issues/403


vtnate (approved):
Good work @amyeallen1 ! I assume it accomplishes your goal, and the python code looks good.

amyeallen1:
Thanks, Nate, for reviewing it! Yes, it does accomplish the goal of making the template extensible to different TEASER building types. Should I go ahead and merge? @vtnate

PR #415, opened 2021-08-04
refactor: add sys params to load partial templates

macintoshpie:
#### What does this PR accomplish?
- Adds building sys params to:
  - load model instance partial templates
  - component definitions partial templates (if coupling includes a load)
  - connect statements partial templates (if coupling includes a load)

Also includes changes from #419 which was merged into this PR's branch. TLDR is we moved some sys params around so that we can also drop them into other couplings.

Also includes changes from #418 which was merged into this PR's branch. TLDR we used the sys params from the district_system in the network 2 pipe <> Heating Plant coupling file

#### How should this be manually tested?
- Edit one of the templates to use `sys_params`, run a test which uses the template, verify it was templated into the resulting model.

#### What are the relevant tickets?
First step into solving #401 

#### Screenshots (if appropriate)
Example templating into a coupling file (⚠️ building sys params have moved since this screenshot, they're now at `sys_params.building`)
Screen Shot 2021-08-04 at 10 29 22 AM

Screen Shot 2021-08-04 at 10 28 56 AM

Example templating into an instance file (⚠️ building sys params have moved since this screenshot, they're now at `sys_params.building`)
Screen Shot 2021-08-04 at 10 30 10 AM

Screen Shot 2021-08-04 at 10 29 47 AM

Example templating with district_system params
Screen Shot 2021-08-13 at 4 47 56 PM

Screen Shot 2021-08-13 at 4 47 38 PM

PR #418, opened 2021-08-10
Water temp template

### LINKED ISSUE #401: Suggested modification for heating hot water supply temperature
To modify the heating hot water supply temperature setpoint, the user needs to modify the value in [this template ](https://github.com/urbanopt/geojson-modelica-translator/blob/develop/geojson_modelica_translator/model_connectors/couplings/templates/Network2Pipe_HeatingPlant/ComponentDefinitions.mopt#L10) as well as the building-level setpoint value in the coupling template for the load model type ([this](https://github.com/urbanopt/geojson-modelica-translator/blob/01c893c0d899c39a213d209c09ab76970802b0a0/geojson_modelica_translator/model_connectors/couplings/templates/Teaser_HeatingIndirect/ComponentDefinitions.mopt#L13) as an example for TEASER). It would be nice if these water-side setpoint values for both heating and cooling could be populated based on the system design parameters schema.

vtnate:
Finally de-hardcodes district and load water temp setpoints
Reads district and load water temp setpoints from sys-param file
Combined with #415 , this resolves #401


macintoshpie (approved):
looks good to me! 👍 💯

PR #419, opened 2021-08-11
Schema changes

vtnate:
This addresses points 1 & 2 of [this comment](https://github.com/urbanopt/geojson-modelica-translator/issues/401#issuecomment-896323096) from #401 by organizing our sys-params slightly differently
- Refactors templates to pull sys-param data that is only central-plant related from the district block of the sys-param file
- Adds district blocks to test sys-param files (it's possible I added district blocks in a few places where they're not needed. I don't think that harms anything, if there were some extra json lines)
- Removes outdated params from schema and test sys-param files
`poetry run py.test tests/model_connectors/test_district_heating_system.py` should return a different value for `T_a1_nominal` in the `Substation/HeatingIndirect_<id>.mo` file when you change the `temp_setpoint_hhw` parameter in `time_series_system_params_ets.json`
#### Other thoughts
There are district params that are still in the ETS section of sys-params. @amyeallen1 confirmed that the remaining `..._district` parameters are appropriate in the ETS block.


geojson_modelica_translator/model_connectors/model_base.py
```diff
@@ -78,6 +78,17 @@ def __init__(self, system_parameters, template_dir):
...
+        if system_parameters is not None:
+            self.district_template_data = {
+                "temp_setpoint_hhw": self.system_parameters.get_param(
+                    "$.district_system.default.central_heating_plant_parameters.temp_setpoint_hhw"
+                ) + 273.15,
+                "temp_setpoint_chw": self.system_parameters.get_param(
+                    "$.district_system.default.central_cooling_plant_parameters.temp_setpoint_chw"
+                ) + 273.15,
+            }
+
```
macintoshpie, line 91: Niiiice! I like that this is at the base model 😎

macintoshpie (approved):
Looks good to me!

geojson_modelica_translator/model_connectors/model_base.py
vtnate: The same variables are being read in the heating and cooling plants, but I think it's ok to read them in twice because this for an ETS not a central plant. Does that sound right?

PR #439, opened 2022-02-10
Debug/sys params

macintoshpie:
GMT developers would like to know which system parameters are/are not being propagated into Modelica templates.

- Adds a poetry command which will iterate through an example system parameters file, tweak them one at a time, and see if it generates any differences in the package.
- runs this poetry command during CI -- note that the command makes no assertions so it will pass every time (even if there are parameters which are unused) -- it's purely for informational purposes currently

Feel free to run the command locally, `poetry run check_sys_params`

https://github.com/urbanopt/urbanopt-modelica/issues/118


pyproject.toml
```diff
@@ -63,3 +63,4 @@ uo_des = 'management.uo_des:cli'
 format_modelica_files = 'management.format_modelica_files:fmt_modelica_files'
 update_schemas = 'management.update_schemas:update_schemas'
 update_licenses = 'management.update_licenses:update_licenses'
+check_sys_params = 'management.check_sys_params:check_sys_params'
```
nllong, line 66: 😻

nllong (approved):
awesome sauce!

vtnate (approved):
Ran it manually. Takes a long time to go through everything, but appears to do what it says on the tin. Love it.

management/data/baseline_sys_params.json
```diff
@@ -0,0 +1,148 @@
+{
+  "buildings": {
+    "default": {
+      "load_model": "time_series",
+      "ets_model": "Indirect Heating and Cooling",
+      "ets_model_parameters": {
```
macintoshpie, line 6: @nllong @vtnate just FYI -- this baseline sys params file is how the check_sys_params knows what sys params to test. Therefore, if it's missing anything we think it should be checking, we should add it to this file!

management/check_sys_params.py
```diff
@@ -0,0 +1,278 @@
...
+        # for each param, see what happens to the package when we modify it
+        for idx, param in enumerate(params):
+            print(DELIMITER, file=output_file_handle)
+            if param['skip']:
+                print(f'{param["path"]} Skipped (duplicate)', file=output_file_handle)
+                continue
+
+            if param['type'] == 'string':
+                print(f'{param["path"]} Skipped (string)', file=output_file_handle)
+                continue
```
macintoshpie, line 249: we skip string parameters for now b/c it's a bit more complicated to guess what the valid enums are, and how they should affect other properties in the sys params file

PR #447, opened 2022-03-09
change mbl_v8 option to be mbl_v9/master

vtnate:
Master branch of MBL is soon going to be v9. We had been using v8.1.0 for "modern" mbl in the gmt tests.

##### _This breaks tests/GMT_Lib/test_gmt_lib.py::test_simulate_cooling_plant_
Cooling plant changed in MBL v8 -> master which breaks this test.

Changes the option for testing with modern mbl to use version 9/master instead of version 8.1
py.test


nllong:
I got a bit closer but I think we will need to chat with Chengnan. The MBL library between 8.1 and 9.0 changed a bit... maybe just paths, but not sure how to resolve. Also, I'm wondering if JModelica will work with MBL 9.0 which has MSL 4.0. Maybe that is the issue?

nllong (approved):
This should pass now! I've added a new simulation mark for `msl_v4_simulation` so that we can run mbl v9 tests with msl v3.2.3 if we need to.

nllong:
doh, looks like this test fails!

`tests/model_connectors/test_time_series_ets.py::TimeSeriesETSTest::test_time_series_ets`

any thoughts?

vtnate:
> doh, looks like this test fails!
> 
> `tests/model_connectors/test_time_series_ets.py::TimeSeriesETSTest::test_time_series_ets`
> 
> any thoughts?

Yeah, the bottom of that test still had the bit where it ran the model with JModelica, which as we know won't work. I've commented that block out, we should be good to go now.

nllong:
> > doh, looks like this test fails!
> > `tests/model_connectors/test_time_series_ets.py::TimeSeriesETSTest::test_time_series_ets`
> > any thoughts?
> 
> Yeah, the bottom of that test still had the bit where it ran the model with JModelica, which as we know won't work. I've commented that block out, we should be good to go now.

Oh, so TimeSeries ETS are now MSL4.0/MBL9.0, right? Because of the integrated ETS with the Building?

vtnate:
That test you mentioned is a new test for the combined TimeSeries_with_ets. The existing time_series tests haven't changed.

nllong:
cool cool! I'll let you merge this down when you are ready!

PR #448, opened 2022-03-09
Clarify c to k

vtnate:
Tryna make things a little better every day
Fun fact: I wrote the function name, github copilot wrote everything else. Wheee, AI!
Replaces `+ 273.15` around the gmt with a utility method that does the same thing
py.test


vtnate:
Since I'm screwing around in util.py, maybe we should move [this method](https://github.com/urbanopt/geojson-modelica-translator/blob/develop/geojson_modelica_translator/model_connectors/model_base.py#L100) in there as well?

nllong (approved):
nice clean up!

PR #452, opened 2022-04-25
upgrade poetry dependencies

nllong:
Simply ran `poetry update`. I like that babel removed support for all python 2*! Onward!

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -50,11 +50,11 @@ toml = "*"
 
 [[package]]
 name = "babel"
-version = "2.9.1"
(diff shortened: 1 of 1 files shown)


vtnate:
A few dependencies in pyproject.toml have minimum versions, while most have exact versions. Is it too risky to allow upgrades?

nllong:
If they are pinned to a specific version, then yeah. I typical pin and say something when I notice an issue. These updates were, I think, innocuous.

vtnate (approved):
One less thing that deals with 2.7!

PR #467, opened 2022-05-17
Microgrid lines

### LINKED ISSUE #466: Get ACLine template working
### Expected Behavior
ACLine.mot creates a functional model

### Actual Behavior
- Wires aren't being properly inserted into the modelica template
- microgrid sys-param test file is outdated, should be using the modern output from opendss

### To resolve:
- [x] Regenerate microgrid sys-param file for tests to use
- [x] Map from [opendss/uo-sdk output](https://github.com/urbanopt/geojson-modelica-translator/blob/microgrid-lines/tests/data_shared/system_params_microgrid_example.json#L353-L358) to [MBL wires](https://github.com/lbl-srg/modelica-buildings/tree/master/Buildings/Electrical/Transmission/MediumVoltageCables) which should be possible using [these segments](https://github.com/urbanopt/urbanopt-ditto-reader/blob/develop/example/extended_catalog.json#L14985-L14993) of the ditto-reader extended catalog.
- [x] Confirm/update use of `perLin` in the template to combine wires into a Line

### Steps to Reproduce
- `poetry run py.test tests/GMT_Lib/test_gmt_lib.py`
- Open ACLine<x>.mo model in Dymola
- Simulate model
- View errors

### Version Information

MBL: v9
GMT: v0.2.3
This is new work that has never worked properly. These versions are not the cause of the error but proof that it doesn't work even on these modern versions.

Notes: hiccup is in system params=>Modelica step

Close to getting resolved, with help from Mingzhe

Working on mapping between MBL and DiTTO reader

vtnate:
The ACLine.mot template for microgrid power lines doesn't handle the wires properly. This fixes it so the template builds a model the way I expect.
- [x] Update microgrid sys-param test file with modern output from opendss
- [x] Fixes bugs in build-sys-param code that crashed if reopt data wasn't available
- [x] Maps from OpenDSS output to MBL wires
- [x] Adds wires to the line properly
1. `poetry run pytest tests/GMT_Lib/test_gmt_lib.py -m 'not simulation'`
2. Load Dymola with MSLv4, MBLv9, and tests/GMT_Lib/output/DistributionLines/ACLine*.mo
3. Simulate the model in Dymola
4. Inspect output of simulation
Resolves #466


amyeallen1 (commented):
Hi Nate, 
     I tried out the test, and it works perfectly to generate the model package, which then simulates without errors.  I simulated the model, and also the ACLineMedium model from the buildings library, and there were some elements of the results that seemed unexpected to me, but they were the same in both cases. It's possible that I'm misinterpreting. Here were those two items that I noted about the model results , for both this model and the original one in the MBL, (which I don't think have anything to do with your PR, so this is tangential): 

1. There's no power dissipated thru any of the resistors, despite there being current thru the line, and a non-zero resistance value assigned. 
image

nllong:
> Hi Nate, I tried out the test, and it works perfectly to generate the model package, which then simulates without errors. I simulated the model, and also the ACLineMedium model from the buildings library, and there were some elements of the results that seemed unexpected to me, but they were the same in both cases. It's possible that I'm misinterpreting, or using different model inputs would generate more expected results. Here were those two items that I noted about the model results , for both this model and the original one in the MBL, (which I don't think have anything to do with your PR, so this is tangential):
> 
> 1. There's no power dissipated thru any of the resistors, despite there being current thru the line, and a non-zero resistance value assigned.
>    image
> 2. The phase angle vector (in radians, according to the documentation) appears to be being reported cumulatively, and as a scalar:
>    image
> 
> Nate and Nick, do either of you have any ideas on what might explain these two things? Thanks! @vtnate @nllong

Can you plot the current and voltage too? This does seem strange, but it is good that MBL and GMT are resulting in the same. The cumulative phase angle is also strange... should be instantaneous.

amyeallen1:
Hi Nick, and Nate, 
      Thanks for considering this. Here are plots of current at port n (the other port is generally equal and opposite). These two lines are representative of the others. 

image

image

These were representative of the voltage values in the lines, with the other element of the voltage vector at the node being zero. 

image

amyeallen1 (approved):
Based on Mingzhe's comments, it appears that the model performed as expected. : )

PR #472, opened 2022-05-24
Python version

vtnate:
Excludes python 3.10 from list of supported Python versions (printed on pypi)

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -81,11 +81,11 @@ six = ">=1.11"
 
 [[package]]
 name = "certifi"
-version = "2021.10.8"
+version = "2022.5.18.1"
 description = "Python package for providing Mozilla's CA Bundle."
 category = "main"
 optional = false
-python-versions = "*"
+python-versions = ">=3.6"
 
 [[package]]
 name = "cfgv"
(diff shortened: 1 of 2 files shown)


nllong:
how about adding python 3.9 to the ci world? Add here: https://github.com/urbanopt/geojson-modelica-translator/blob/develop/.github/workflows/ci.yml#L21

vtnate:
@nllong Do we have any constraints on CI? I just tripled the number of checks with this PR 😬

.github/workflows/ci.yml
```diff
@@ -18,11 +18,18 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
-        # 3.9 does not work as of 2020-02-01
-        python-version: [3.7]
+        python-version: [3.9, 3.8, 3.7]
         test_env: [python, precommit, docs]
         mbl_branch: [master]
         include:
+          - os: ubuntu-latest
```
nllong, line 25: I don't think you need to redefine these includes for each python version. Just replace the 3.7 with `${{ matrix.python-version }}`. Again, I think, haha.

nllong: actually, i think you can just delete the python-version line altogether in the `include` block.

vtnate: Sweet, that went swimmingly. We're good to go.

PR #477, opened 2022-06-07
bump dependency versions

nllong:
Bump locked versions


pyproject.toml
```diff
@@ -28,32 +28,31 @@ packages = [
...
-pre-commit = "2.7.1"
-pytest = "6.1.0"
-pytest-cov = "2.11.1"
-sphinx = "4.5.0"
-sphinx_rtd_theme = "0.5.0"
-sphinx-jsonschema = "1.16.7"
-syrupy = "^1.5.0"
+autopep8 = "1.6.0"
+coveralls = "3.3.1"
+#flake8 = ">=4.0.1"
```
vtnate, line 47: Is this supposed to be commented?

pyproject.toml
```diff
@@ -28,32 +28,31 @@ packages = [
...
+mypy = "^0.961"
+pre-commit = "2.19.0"
+pytest = "7.1.2"
+pytest-cov = "3.0.0"
+sphinx = "5.0.1"
```
vtnate, line 52: Has this changed to esbonio?

vtnate (changes requested):
Besides my 2 small questions, LGTM.

pyproject.toml
nllong: yeah, i just removed it. We use the flake8 that is in pre-commit now.

pyproject.toml
nllong: I don't think so. Esbonio looks like a language server for sphinx? https://pypi.org/project/esbonio/.. says early development too.

vtnate (approved):

PR #480, opened 2022-06-09
compile & run with spawn & optimica

### LINKED ISSUE #423: Integrate ability to run GMT using Spawn compiler
Will be tied to MBL master branch which is Modelica 4.0.

JModelica is being slowly deprecated. Will primarily rely on Optimica. 

Has requested mod to AOP to buy license for everyone who needs to compile from MBL and limited amount of their own content, relative to # of MBL/MSL files (probably done in advance of 9/30)

License will have an indefinite term, but won't include enhancements; implicitly tied to an MSL version 

Updates to MBL: Modelon will provide a webservice that will handle changes to MBL changes that could affect compiler ("signing service"), based on commit to MBL (Kyle has tested a one off case)

Nate has a PR in progress

Limitations with running a Docker container  on a Mac M1 (due to container version)

On a different computer, very close to being done

Near term solution: continue to use local Modelon license. Longer-term solution is OpenModelica. Maybe available by Q1 of next FY. 

`spawn-optimica` branch.

Now successfully compiling with Spawn, and using "fmu runner" (helper class). Works for example models so far, but not DES, due to permission hiccups. Kyle can address this issue in a custom build.

### LINKED ISSUE #460: How to set Optimica env var when running MBL CI
Kyle will address this with notes/instructions for Nate

vtnate:
Compiling with Spawn is a way of getting access to MSLv4, and therefore MBLv9
- Replaces jmodelica.py with a spawn.py file
- modelica_runner.py calls the new Spawn file for compiling, and uses Spawn to simulate
    - Currently we simulate with Optimica. Spawn will also allow OpenModelica to be used for simulations (hopefully by 2023-01-01)
Resolves #423 Resolves #460


vtnate:
Fun fact: the `concurrency` change in the CI config works! It automatically cancels CI runs when another of the same type is triggered. Hooray efficiency.

PR #498, opened 2022-09-20
Update schema.json

vtnate:
Clarify schema that weather file needs to be in Modelica (*.mos) format


nllong (approved):
Nice. Thanks for updating.

PR #514, opened 2022-12-13
Improve ci config settings

vtnate:
Our CI should test the appropriate settings and environments

#513  failed CI which made this solution apparent. Merging this PR should make CI pass in that PR.
- Explicitly use Ubuntu 20.04 because apparently 22.04 doesn't support Python 3.7
- Add the Github Token secret to make Coveralls happy
- Explicitly point to the MBL 9.1 branch, instead of master
- Adds the message command currently required with pytest to not run compilation or simulation tests


nllong (approved):
yikes, new ubuntu doesn't support python 3.7. That will be a fun upgrade with pyfmi related dependencies.

Thanks for fixing this!

PR #517, opened 2022-12-16
Detailed models

vtnate:
We should handle geojson files that include detailed models
- Move check in geojson.py to the appropriate place
- Remove area from LoadBase output
Run a test using a geojson file that refers to detailed models

### DIFF
--- a/geojson_modelica_translator/geojson/urbanopt_geojson.py
+++ b/geojson_modelica_translator/geojson/urbanopt_geojson.py
@@ -94,18 +94,16 @@ class UrbanOptGeoJson(object):
         self.buildings = []
         for feature in self.data.features:
             if feature["properties"]["type"] == "Building":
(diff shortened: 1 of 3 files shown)


nllong:
per your question... are any of the inherited base models using the building area? If not, then feel free to demote it. otherwise, we should respect the inheritance (I think). 

Thanks for finding this and supporting our users!

vtnate:
`area` appears to go to here and stops. https://github.com/urbanopt/geojson-modelica-translator/blob/develop/geojson_modelica_translator/model_connectors/load_connectors/teaser.py#L134. I don't like that line, setting `net_leased_area = building_area` doesn't seem like the right thing to do. If we pass `0` (when we don't have a building area) the model will build.

nllong (approved):
looks great!

PR #518, opened 2022-12-16
Python 311

vtnate:
Support Python 3.11. Some of the dependencies that support Python 3.11 require Python 3.8.
- Removes support for python 3.7
- Adds support for python 3.11
- Adds message to tox env to skip compilation and simulation tests
- Update dependency versions
- Install new Python version
- Follow normal installation procedure https://docs.urbanopt.net/geojson-modelica-translator/developer_resources.html#getting-started-as-a-developer
- Expect all green at `poetry run pytest` and `poetry run tox`

~This needs to be confirmed to work with compiling before it is ready to merge. [This test](https://github.com/urbanopt/geojson-modelica-translator/blob/develop/tests/modelica/test_modelica_runner.py#L124-L146) needs to pass.~

The compilation requirement is dropped temporarily. We depend on Dymola for compilation and simulation until a suitable solution is ready, probably OpenModelica.


nllong:
@vtnate  - can you resolve these conflicts?

nllong (approved):
looks great!

PR #507, opened 2022-11-29
Cli bugfixing

vtnate:
#### Any background context you want to provide?
A Windows user reported problems with the CLI in the attached issue. This fixes that concern.
#### What does this PR accomplish?
- Removes explicit casting to a float
- Adds a log entry
- Updates dependencies
- Makes mos weatherfile variable name more explicit
- Updates geojson file used in test to match the SDK
#### How should this be manually tested?
I'd like a human Windows user (not just our Windows CI) to confirm this passes:
`poetry run pytest tests/management/test_uo_des.py::CLIIntegrationTest::test_cli_builds_sys_params`
#### What are the relevant tickets?
Resolves #503

PR #519, opened 2022-12-28
bugfix of the bugfix, yay

vtnate:
I removed a duplicate casting to float in #507 but it turns out there are cases where the second casting is required. My current theory is that `building_type: "lodging"` in the geojson file means `massFlowRateHeating` for the entire building is zero, even in a Mixed Use building if lodging is a non-zero percentage of the building. This zero, in turn, broke system_parameters.py by being an int and not a float.
Force casting to float when mfrt heating load is 0


nllong (approved):
nice ... could have just been `float()` or not?

vtnate:
> nice ... could have just been `float()` or not?

Lol. Yes, it can. I'll do that.

PR #522, opened 2023-01-05
Remove support for 3.7 in development, update development dependencies

nllong:
Versions of flake8 and new BuildingsPy now require python 3.8

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,33 +29,33 @@ packages = [
 
 [tool.poetry.dependencies]
 # Urbanopt SDK requires Python 3.10 as of UOv0.9.0
-python = ">=3.7.1, <3.11"
-BuildingsPy = "3.0.0"
+python = ">=3.8, <3.11"
+BuildingsPy = "4.0.0"
 click = "8.1.3"
 geojson = "2.5.0"
 jinja2 = "3.1.2"
 jsonpath-ng = "1.5.3"
-jsonschema = "4.6.0"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -29,33 +29,33 @@ packages = [
...
-python = ">=3.7.1, <3.11"
-BuildingsPy = "3.0.0"
+python = ">=3.8, <3.11"
+BuildingsPy = "4.0.0"
 click = "8.1.3"
 geojson = "2.5.0"
 jinja2 = "3.1.2"
 jsonpath-ng = "1.5.3"
-jsonschema = "4.6.0"
+jsonschema = "~4.17"
```
vtnate, line 38: ❤️  tilde versioning

pyproject.toml
```diff
@@ -29,33 +29,33 @@ packages = [
 
 [tool.poetry.dependencies]
 # Urbanopt SDK requires Python 3.10 as of UOv0.9.0
-python = ">=3.7.1, <3.11"
-BuildingsPy = "3.0.0"
+python = ">=3.8, <3.11"
```
vtnate, line 32: If we don't have to support 3.7, I think we can then support 3.11.

vtnate (approved):
Out with the old, in with the new!

pyproject.toml
nllong: yeah, it's my favorite. So much better than `>`

PR #525, opened 2023-01-09
specify ports for each time series building in instance template

vtnate:
We found our time-series instance template was missing information about the available ports
Specifies that there are one each of input/output ports for hot and cold water


nllong (approved):
nice catch. I assume that these were working without the port declarations and this is being just more explicit?

vtnate:
This is a fix for some of the issues that arose when preparing for the DOE demo. I'm having trouble linking to the Teams message.

PR #530, opened 2023-03-07
add max_electrical_load to building sys-param data

vtnate:
We need a nominal building load for microgrid power source models to work properly. We can grab it from OpenStudio output
This reads the OpenStudio simulation output via the  export_modelica_loads measure to find the max FacilityElectricity value and adds it to the sys-param file generated by the CLI
Generate your own OpenStudio data (using the SDK CLI), and make a sys-param file using the DES CLI on this branch.


geojson_modelica_translator/system_parameters/time_series_microgrid_template.json
```diff
@@ -17,7 +17,8 @@
           "has_liquid_cooling": true,
           "has_electric_cooling": false,
           "temp_chw_supply": 7,
-          "temp_chw_return": 12
+          "temp_chw_return": 12,
+          "max_electrical_load": 0
```
nllong, line 21: any reason to alphabetize this?

nllong (approved):
lit!

geojson_modelica_translator/system_parameters/time_series_microgrid_template.json
vtnate: No need to, but that's a good idea. I'll alphabetize and merge.

PR #534, opened 2023-04-11
update dependency versions

nllong:
Needed to update some dependency versions for the DC analysis

Simply updates the pinned versions of the dependencies

CI should suffice

n/a


vtnate (approved):
Some primary dependencies are fixed, some are tilde, some are caret. Are those being specifically used, or should we eventually set them to all tilde like the dev-dependencies?

nllong:
i only `~` the most stable-type dependencies. So basically anything we don't write (hehe) or ones that we have had problems with.

PR #535, opened 2023-04-12
Removing remaining tox calls

nllong:
We removed tox from the CI and it should have been removed from the rest of the code.

* remove tox library
* update docs, code, etc to support change

If CI passes, we should be good.

n/a


vtnate (approved):
I love PRs that remove more code than they add! 😆 ❤️

Issue #528, opened 2022-12-07
Modify System Parameter Schema to allow non-geometric user inputs


JingWang-CUB:
This issue is to modify the GeoJSON Schema to allow users to define non-geometric inputs related to GHE networks, including:

- Max/min fluid temperatures
- Pipe sizes
- Thermal properties
- Others TBD

PR #537, opened 2023-04-12
Modify GHE params

tanushree04:

Addresses changes for GHE 

This PR adds schema changes for GHE, adds example files and tests.

The example file to be used for GHE is:  `tests/system_parameters/data/system_params_ghe.json`

poetry run py.test tests/system_parameters/test_system_parameter.py

#528


tanushree04:
The following error on the CI `ValueError: The type of geothermal heat exchanger pipe arrangement is not supported.` is to be expected with updating the ghe parameters in the schema and will be resolved once this PR is merged

tests/model_connectors/data/system_params_ghe.json
```diff
@@ -107,32 +107,32 @@
...
+        "pipe": {
+          "inner_diameter": 0.0216,
+          "outer_diameter": 0.0266,
+          "shank_spacing": 0.0323,
+          "roughness": 1e-06,
+          "conductivity": 0.4,
+          "rho_cp": 1542000
+        },
+        "borehole": {
+          "length": 96.0,
```
mitchute, line 131: What does this length map to? Does it map directly to the already-sized GHE or something else?

tests/system_parameters/data/system_params_ghe.json
```diff
@@ -2,137 +2,8843 @@
...
-        "heat_flow_nominal": 8001,
-        "mass_hhw_flow_nominal": 11,
-        "boiler_water_flow_minimum": 11,
-        "pressure_drop_hhw_nominal": 55001,
-        "pressure_drop_setpoint": 50000,
-        "temp_setpoint_hhw": 54,
-        "pressure_drop_hhw_valve_nominal": 6001,
-        "chp_installed": false
+      "ghe_parameters": {
+        "version": "0.5",
```
mitchute, line 34: It won't make any difference, byt GHEDesigner is on v1.0

mitchute (commented):
Thanks, @tanushree04. I made a few comments for you to consider.

tests/system_parameters/data/system_params_ghe.json
vtnate, line 1: This test file looks great. Can we make the test sys-param files in model_connectors look like this as well, or are they testing different capabilities?

vtnate (commented):
LGTM, besides my one comment on the test sys-param files. Let's see what the others have to say.

tests/system_parameters/data/system_params_ghe.json
tanushree04: @JingWang-CUB do you have any thoughts on this?

tests/model_connectors/data/system_params_ghe_2.json
```diff
@@ -204,8822 +204,8822 @@
...
+          "method": "rectangular"
+        },
+        "design": {
+          "flow_rate": 0.2,
+          "flow_type": "borehole",
+          "max_eft": 35.0,
+          "min_eft": 5.0,
+          "number_of_boreholes": 4
+        },
+        "placeholder": {
```
JingWang-CUB, line 256: These placeholders can be kept for now. I will delete them later. But I think we agreed to add scenario_dir or ghedesigner_dir to the schema? See below for my notes. 
image

tests/model_connectors/data/system_params_ghe_3.json
JingWang-CUB, line 1: Same comment as above.

JingWang-CUB (commented):
Hi @tanushree04 , I left some comments for your consideration. Let me know if you have any questions. Thanks!

tests/model_connectors/data/system_params_ghe.json
tanushree04: @JingWang-CUB are you adding this from the GHE sizing result?

JingWang-CUB: My understanding is that this will be updated by the GHEDesigner sizing outputs.

tests/model_connectors/data/system_params_ghe.json
```diff
@@ -2,117 +2,44 @@
...
-        "pressure_drop_hhw_valve_nominal": 6001,
-        "chp_installed": false
-      },
       "ghe_parameters": {
-        "connectors": {
-          "pipe_diameter": 12,
-          "pipe_insulation_rvalue": 5,
-          "pipe_location": "Tunnel"
+        "version": "1.0",
+        "ghe_dir": "C:/urbanopt-ghe/baseline_scenario/ghe",
```
mitchute, line 35: Does this path matter?

tanushree04: @mitchute this points to the GHE sizing results folder within the URBANopt scenario directory. This folder will be created when GHEDesigner is called from the CLI. GMT will read results from this folder to run analysis.

mitchute: Thanks @tanushree04. I guess my question is more along the lines of that since this appears to be inside a test, whether the Windows path is correct. If the tests all pass and this is correct, feel free to ignore my comment.

tanushree04: Its just validating whether the path is a string in the schema as of now.

tanushree04:
@mitchute and @JingWang-CUB I have added the fifth gen ets parameters and added the geometric properties for GHE as an array

tests/system_parameters/data/system_params_ghe.json
```diff
@@ -24,7 +24,17 @@
       },
       "ets_model": "Fifth Gen Heat Pump",
       "fifth_gen_ets_parameters": {
-        "supply_water_temperature_building": 15
+        "supply_water_temperature_building": 15,
+        "chilled_water_supply_temp": 5,
+        "hot_water_supply_temp": 50,
```
vtnate: Does this mean the temps supplied to the building from the ETS heat pump?

vtnate:
The current borefield test failure is different from the `ValueError` originally noted. Is that expected?

tanushree04:
> The current borefield test failure is different from the `ValueError` originally noted. Is that expected?

Its expecting to find the temp_setpoint_hhw if it is a fifth_generation district system. This line is failing: 
https://github.com/urbanopt/geojson-modelica-translator/blob/bee5e9e3f76485fba8595896b5531fdec29b2c20/geojson_modelica_translator/model_connectors/model_base.py#L62

@vtnate @JingWang-CUB do we need to modify this after this PR is merged?

tests/model_connectors/data/system_params_ghe.json
```diff
@@ -2,137 +2,8865 @@
...
+          "shank_spacing": 0.0323,
+          "roughness": 1e-06,
+          "conductivity": 0.4,
+          "rho_cp": 1542000,
+          "arrangement": "singleutube"
+        },
+        "borehole": {
+          "length": 96.0,
+          "buried_depth": 2.0,
+          "diameter": 0.15
```
JingWang-CUB, line 71: Do we need to have a default number_of_boreholes defined in this initial file and then update it?

tests/system_parameters/data/system_params_ghe.json
JingWang-CUB: I think they should be exactly the same set of sys-param files. Feel free to overwrite the ones in tests/model_connectors with the updated ones here. The changes will expectedly break my test. I plan to fix them once this PR is merged.

tests/system_parameters/data/system_params_ghe.json
```diff
@@ -2,137 +2,8864 @@
...
+          "outer_diameter": 0.0266,
+          "shank_spacing": 0.0323,
+          "roughness": 1e-06,
+          "conductivity": 0.4,
+          "rho_cp": 1542000,
+          "arrangement": "singleutube"
+        },
+        "borehole": {
+          "buried_depth": 2.0,
+          "diameter": 0.15
```
JingWang-CUB: Same as above, should we have a default number_of_boreholes here?

tests/model_connectors/data/system_params_ghe.json
tanushree04: @mitchute do you have thoughts on this

tanushree04: Same goes for the `length` of the borehole

mitchute: Will there also be default g-function values? If we have one, I think we should have the other(s) with corresponding values.

tests/system_parameters/data/system_params_ghe.json
mitchute: @JingWang-CUB @tanushree04 same comment as above.

tests/model_connectors/data/system_params_ghe.json
JingWang-CUB: Currently, we only specify the path to the g-function file directory. I guess we could specify a default g-function value in this file in case the generation of the g-function from previous steps didn't go through.

tanushree04: Since the number_of_boreholes are updated after the GHE Designer run I have not listed that as a required property in the schema. Same goes for the length of boreholes. Since Chris needs an example file that is going to be an input for the GHE designer, I believe it should be okay to share one which doesnt have the number of boreholes.

mitchute:
@tanushree04 thanks for the work here. It looks OK to me.

vtnate (approved):

JingWang-CUB (approved):

PR #543, opened 2023-04-28
GMT parameter transfer files for electrical components

ZhanweiHe:
GMT parameter transfer files for generator, grid, battery, capacitive load, inductive load, transformer, and converter.

### DIFF
--- a/geojson_modelica_translator/modelica/GMT_Lib/Electrical/AC/ThreePhasesBalanced/Conversion/ACACTransformer.mot
+++ b/geojson_modelica_translator/modelica/GMT_Lib/Electrical/AC/ThreePhasesBalanced/Conversion/ACACTransformer.mot
@@ -1,0 +1,41 @@
+within{{project_name}}.ACACTransformer;
+model {{data['model_name']}}
+  "Isolated transformer template model for GMT level 1 testing"
+  parameter Modelica.Units.SI.Voltage VHigh = {{data["RMS voltage on high side"]}} "Rms voltage on side 1 of the transformer (primary side)";
+  parameter Modelica.Units.SI.Voltage VLow = {{data["RMS voltage low side"]}} "Rms voltage on side 2 of the transformer (secondary side)";
+  parameter Modelica.Units.SI.ApparentPower VABase = {{data["Nominal capacity"]}} "Nominal power of the transformer";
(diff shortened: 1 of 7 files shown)


geojson_modelica_translator/modelica/GMT_Lib/Electrical/AC/ThreePhasesBalanced/Conversion/ACACTransformer.mot
```diff
@@ -0,0 +1,41 @@
+within{{project_name}}.ACACTransformer;
+model {{data['model_name']}}
+  "Isolated transformer template model for GMT level 1 testing"
+  parameter Modelica.Units.SI.Voltage VHigh = {{data["RMS voltage on high side"]}} "Rms voltage on side 1 of the transformer (primary side)";
```
vtnate, line 4: We don't have high-side voltage available. See https://docs.google.com/spreadsheets/d/1LHl3iaCbn5cZQt2V28gTz33ga5Y85dZQ/edit#gid=1654231185 and look in the sections for both Transformer and Substation.

vtnate: I'm researching to understand if it's possible to provide VHigh some other way. Is it possible to make the model work if I can't find a way to provide that value?

ZhanweiHe: Hello Nathan, 
How about we do not transfer the "VHigh" parameter for "high-side voltage"? The model extends from the existing example model in MBL, which has a default value for this parameter. In the example model, the default value of VHigh is 480 V.

vtnate: That sounds great. Lets do that. Thanks!

ZhanweiHe:
@vtnate Nathan, thank you for letting me about this. To clarify, should we stick with the current version for now and plan to add the parameter "VHigh" at a later date?

vtnate:
This work is being used in #549, so this PR is no longer needed.

PR #545, opened 2023-05-03
Enable compile & simulate with OpenModelica in Docker

vtnate:
- We haven't compiled or simulated our models "in-house" in a while. Now we can.
- CI hasn't been changed yet to actually test anything in this PR. This PR is not yet in a state where compiling or simulating will work.
- Builds a Linux image with OpenModelica installed
- Re-activates the shell script to load/activate the container with that image
- Re-activates the under-the-hood cli for calling OM inside the container
Run the test suite locally, except this time DON'T include `-m "not compilation and not simulation"` 😱


vtnate:
FYI: Several of the CI checks failed on the `Install dependencies with Poetry` step. I kept re-running the failed jobs until they passed, so no problems with our code. Each re-run failed at poetry installing a different dependency, which makes me a little uncomfortable with poetry.

No action necessary now, just my $.02

PR #548, opened 2023-05-05
Add MOS file wrapper and size the 5G mass flow rate

nllong:
We were parsing the MOS file by hand, the sizing of the mass flow rate was fixed, and the plant flow rate.

* Add a new class to read in an MOS file. The class can read the variables in the header and can set variable values.
* Refactor the reading of the water heating load to use the new class
* Expose the plant mass flow rate to the DES5G mot file
* Size the mass flow rates according to the total heating/cooling load, delta T, and Cp

* Ideally, generate the DES_5G test and run in Dymola (I confirmed that it ran)
* The unit tests should cover the MOS file class

n/a

### DIFF
--- a/tests/modelica/test_modelica_mos.py
+++ b/tests/modelica/test_modelica_mos.py
@@ -1,0 +1,94 @@
+# :copyright (c) URBANopt, Alliance for Sustainable Energy, LLC, and other contributors.
+# See also https://github.com/urbanopt/geojson-modelica-translator/blob/develop/LICENSE.md
+
+import filecmp
+import os
+import unittest
+from pathlib import Path
+from geojson_modelica_translator.modelica.modelica_mos_file import ModelicaMOS
+
+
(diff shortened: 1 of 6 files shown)


tests/modelica/test_modelica_mos.py
```diff
@@ -0,0 +1,94 @@
...
+# See also https://github.com/urbanopt/geojson-modelica-translator/blob/develop/LICENSE.md
+
+import filecmp
+import os
+import unittest
+from pathlib import Path
+from geojson_modelica_translator.modelica.modelica_mos_file import ModelicaMOS
+
+
+class ModelicaMOSTest(unittest.TestCase):
```
vtnate, line 11: 😍

vtnate (approved):
The model even compiles and simulates in OpenModelica! 🙌  OM whines a little bit about initial conditions, but it does that with most of our models anyway.

PR #549, opened 2023-05-08
Enhance and expand microgrid templates and code

vtnate:
More templates were required to model the components of a micrigrid. This adds them
New templates and the code to render them for:
- AC inductive loads
- Electrical grids
- Generators
- Batteries
- Capacitors
1. Run the test locally - `poetry run pytest tests/GMT_Lib/test_gmt_lib.py -m "not compilation and not simulation"`
2. Load the resulting models (found in tests/GMT_Lib/output/) in OpenModelica and simulate
3. Expect successful simulations
    1. The `inductive load` models have bad input from UO SDK, so that model is not expected to simulate successfully


vtnate:
Github seems to think my module name (inductive.py) is capitalized, but it isn't in my local version (which git accepts as current). I don't know how to make Github see that it's version is not what my local version is, that I just pushed. Argh!

amyeallen1:
Hey Nate, I can review this later in the week (Thursday or Friday). If a review is needed sooner, maybe Jiazhen or Nicholas can do that. Thanks! @vtnate

amyeallen1:
Tests ran successfully and grid model ran successfully (I simulated for a 1-hour period). Great work! @vtnate

amyeallen1 (approved):

Issue #511, opened 2022-12-07
Update GMT to accept GHE parameters and network configurations


JingWang-CUB:
This issue is for creating new templates and tests for 5G district energy systems with geothermal heat exchangers under the URBANopt GHP project.

JingWang-CUB:
@vtnate , could you please create a new branch based on the current develop branch named issue511_ghe_templates for me to put my new templates and stuff? Thanks!!

vtnate:
@JingWang-CUB Done! https://github.com/urbanopt/geojson-modelica-translator/tree/issue511_ghe_templates

vtnate:
Closed with #533

PR #550, opened 2023-05-17
Fixing borefield test breaks due to schema changes

JingWang-CUB:
URBANopt GHP project

Fix GHE test breaks after system parameter file schema changes

poetry run pytest -m 'not compilation and not simulation' tests/model_connectors/test_borefield.py

#511


vtnate (changes requested):
Looks like `tests/model_connectors/test_borefield.py::DistrictSystemTest::test_build_district_system` is still looking for `tests/model_connectors/data/system_params_ghe_2.json` to use in the test, but it was deleted in this PR. Probably just need to change the test to point at the correct file.

JingWang-CUB:
@vtnate , right, I'm working on a few other changes including modifying the test to point to the correct file.

vtnate:
@JingWang-CUB This is not urgent. Sometime before you merge this, can you change the name to describe what is happening in the PR? Let's start with an action verb too (like `Fixing GHE tests` or `Adding functionality to GHE` or `Removing redundant GHE test files`).

JingWang-CUB:
@vtnate , the PR is ready for your review. The test now can pass and the model can simulate. I also included two temporary changes to suppress the known errors as we took out central heating plant parameters from the GHE system parameters file. Let me know if you have a better way to address those. Thanks!

tests/model_connectors/data/system_params_ghe.json
```diff
@@ -42,7 +42,7 @@
     "fifth_generation": {
       "ghe_parameters": {
         "version": "1.0",
-        "ghe_dir": "C:/urbanopt-ghe/baseline_scenario/ghe",
+        "ghe_dir": "D:/GitHub/geojson-modelica-translator/tests/model_connectors/data",
```
vtnate, line 45: CI can't find this file. Looks like it's an absolute path to somewhere on your computer, and when CI runs it isn't looking on your computer.

geojson_modelica_translator/model_connectors/networks/templates/NetworkAmbientWaterStub_Instance.mopt
```diff
@@ -5,7 +5,7 @@
     redeclare package Medium={{ globals.medium_w }},
     use_m_flow_in=true,
     use_T_in=false,
-    T={{ sys_params.district_system.fifth_generation.central_heating_plant_parameters.temp_setpoint_hhw }}+273.15,
+    T={{ sys_params.district_system.fifth_generation.ghe_parameters.soil.undisturbed_temp}}+273.15,
```
vtnate, line 8: We need to change the original 5G test sys-param file to include `ghe_parameters` instead of `central_heating_plant_parameters` now that you've made this change.

vtnate (changes requested):
Couple things still to change, but this is looking good. Thanks Jing!

tests/model_connectors/data/system_params_ghe.json
JingWang-CUB: Correct. How should I fix this?

vtnate: Make sure the code can handle both relative and absolute paths, and then put a relative path to the test_ghe_dir in this test sys-param file. Thanks!

geojson_modelica_translator/model_connectors/networks/templates/NetworkAmbientWaterStub_Instance.mopt
vtnate: Since we're changing the instance template that the test file uses (and it breaks the test), I think we should change the test file as well.

vtnate:
@JingWang-CUB Try running pre-commit locally, and commit whatever changes it makes. That should make a bunch of these errors go away.

geojson_modelica_translator/model_connectors/networks/templates/NetworkAmbientWaterStub_Instance.mopt
vtnate: As we discussed on the phone, this file needs an if/else to handle the ghe_parameters. Eventually we'll need to change the central_heating_parameters, because that doesn't make sense for a fifth-gen district, but that should be a separate PR from this.

JingWang-CUB: Hi Nate, I just pushed some other commits which fix this by adding an if/else statement. All pre-commit tests have passed. Let's see what the CI tests say.

geojson_modelica_translator/model_connectors/networks/templates/NetworkAmbientWaterStub_Instance.mopt
```diff
@@ -5,7 +5,11 @@
     redeclare package Medium={{ globals.medium_w }},
     use_m_flow_in=true,
     use_T_in=false,
+    {% if sys_params.district_system['fifth_generation']['ghe_parameters'] %}
```
vtnate, line 8: Jinja has a weak if-engine. I think if you reverse this it'll work. Like, `if not ...['ghe_parameters']` then the normal way, else read the ghe_parameters.

vtnate (approved):

PR #551, opened 2023-05-17
Updating GHE Parameters: Modifying GHE Parameters schema and example files

tanushree04:
- Adds ghe specific params to ghe_params properties. This is an array that includes properties for each GHE instance. These include borehole length and number of boreholes (determined from GHE sizing), ground loads (determined from the Thermal Network sizing Calculations) and the geometric constraints (the length and width of the GHE - determined from the GeoJSON Feature file). The borehole length, number of boreholes and ground loads are optional properties.
- Removes loads as a required GHE parameter since these added to the system parameters after calling the Thermal Network from the CLI and passing on the system parameters as an input.


tanushree04:
@mitchute Should the ground loads be an array with associated GHE Feature ID - for cases there are more than 1 GHE in the network? 

https://github.com/urbanopt/geojson-modelica-translator/blob/0a5343e57a9dc9d9e45d6cd7eac99f81a5b501ed/tests/system_parameters/data/system_params_ghe.json#L101

https://github.com/BETSRG/GHEDesigner/blob/ff73e8a93cb3c1a564e1998d69fb6ebbe07a9766/ghedesigner/schemas/loads.schema.json#L4

mitchute:
@tanushree04 where are the unique number of boreholes and depths for each GHE going to go? Should those also be arrays, or do you think we need a separate object that holds the loads, NBH, and depth for each GHE instance?

mitchute:
> Are the depth of the borehole also unique for each GHE ?

Yes, that's correct.

vtnate:
@tanushree04 This is not urgent. Sometime before you merge this, can you change the name to describe what is happening in the PR? Let's start with an action verb too (like `Fixing GHE tests` or `Adding functionality to GHE` or `Removing redundant GHE test files`).

geojson_modelica_translator/system_parameters/schema.json
vtnate, line 1: Something in the test file under `ghe_parameters` doesn't meet the schema definition, which is causing the CI failures.

vtnate (changes requested):
In general this is good. Gotta make the test file meet the schema definition, then we're good.

geojson_modelica_translator/system_parameters/schema.json
tanushree04: I updated the model_connectors examples as well - now we see a  KeyError: 'central_heating_plant_parameters' that needs to be probably fixed in Jing's PR

vtnate: Excellent. Yes, that KeyError is being addressed in Jing's PR

vtnate (approved):

mitchute (approved):

PR #552, opened 2023-06-03
Enable tests to run in docker with omc

nllong:
Since Jmodelica became an unsupported project, we have not been able to run simulations with the GMT. This impacted both the user and the unit/integration tests.

* Re-enable running of the unit tests by packaging openmodelica in a docker container
* Update the signature of the `run_in_docker` command to be more generic. Require an `action`, a `model_name`, the `file_to_load`, and the `run_path`. 
* Update the omc workflow to generate the `mos` simulate and create_fmu files using the Jinja templating
* There are several unit tests that are disabled (search for `pytest.mark.skip`. These tests don't work due to some OMC-related errors that we will need to investigate.

The unit tests should suffice.

n/a


pyproject.toml
```diff
@@ -36,7 +36,7 @@ jinja2 = "3.1.2"
 jsonpath-ng = "1.5.3"
 jsonschema = "~4.17"
 modelica-builder = "^0.2.3"
-pandas = "1.5.2"
+pandas = "2.0.2"
```
vtnate, line 39: Does pandas change enough that going to tilde versioning will bite us? 🤔

tests/model_connectors/test_time_series_mft_ets.py
```diff
@@ -137,17 +143,17 @@ def test_mft_time_series_to_modelica_and_run(self):
...
-        self.assertAlmostEqual(cool_q_flow.max(), 354100, delta=10)
-        self.assertAlmostEqual(cool_q_flow.mean(), 3160, delta=10)
+        self.assertAlmostEqual(cool_q_flow.min(), -131.9, delta=10)
+        self.assertAlmostEqual(cool_q_flow.max(), 512936, delta=10)
+        self.assertAlmostEqual(cool_q_flow.mean(), 30210, delta=10)
 
-        self.assertAlmostEqual(heat_q_flow.min(), -343210, delta=10)
-        self.assertAlmostEqual(heat_q_flow.max(), 39475, delta=10)
-        self.assertAlmostEqual(heat_q_flow.mean(), -23270, delta=10)
+        self.assertAlmostEqual(heat_q_flow.min(), -342201, delta=10)
```
vtnate, line 157: I got a different value for this when testing locally - outside the allowable range so it failed the test. Not sure what, if anything, to do about it.

vtnate (approved):
OMG 🥳 🙌

management/uo_des.py
```diff
@@ -176,7 +176,7 @@ def run_model(modelica_project: Path):
     """
     run_path = Path(modelica_project).resolve()
     project_name = run_path.stem
-    file_to_run = run_path / 'Districts' / 'DistrictEnergySystem.mo'
+    run_path / 'Districts' / 'DistrictEnergySystem.mo'
```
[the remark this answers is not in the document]
nllong: doh, missed this one.

tests/model_connectors/test_time_series_mft_ets.py
nllong: if it happens again then increase the delta...

pyproject.toml
nllong: yeah, let's leave it like this for now

PR #554, opened 2023-06-05
Fix borefield test breaks due to the schema changes in issue #551

JingWang-CUB:
Fix borefield test breaks due to the schema changes in issue #551

poetry run pytest -m 'not compilation and not simulation' tests/model_connectors/test_borefield.py

#551


vtnate (approved):
I love it when fixes are this simple! ❤️

PR #560, opened 2023-06-20
Read correct data from sys-param for microgrid electrical load

### LINKED ISSUE #555: Need new output from SDK for microgrid inductive load
### Expected Behavior
Inductive template to build a model from UO SDK output

### Actual Behavior
[P_nominal](https://github.com/urbanopt/geojson-modelica-translator/blob/microgrid-code/geojson_modelica_translator/modelica/GMT_Lib/Electrical/AC/ThreePhasesBalanced/Loads/Inductive.mot#L4) in the modelica model is expecting a number (the actual power consumption of the load) whereas the UO SDK is outputting [a string](https://github.com/urbanopt/geojson-modelica-translator/blob/microgrid-code/geojson_modelica_translator/system_parameters/schema.json#L1175-L1181). 

Either the SDK needs to output something different, or we need to use a different output of the SDK so the model will work.

vtnate:
The Modelica template for inductive electric load was getting invalid info from the test sys-param file. Now we're reading from the correct section of the sys-param and loading the data properly into the template.
Reads electrical load values for each building to help generate a microgrid
The automated test for `tests/GMT_Lib/test_gmt_lib.py::test_simulate_inductive_load` should be sufficient, but there are still issues with OpenModelica that need to be resolved. Running by hand with Dymola will verify that the template reads data from the sys-param file and properly generates a working model. There is also a test that builds the model, so it can be inspected by hand at `tests/GMT_Lib/output/Inductive/Inductive0.mo`.
Resolves #555

### DIFF
--- a/tests/GMT_Lib/test_gmt_lib.py
+++ b/tests/GMT_Lib/test_gmt_lib.py
@@ -313,6 +313,30 @@ def test_build_inductive_load():
     # Did the mofile get created?
     assert linecount(package_output_dir / 'Inductive0.mo') > 20
 
+
+@pytest.mark.simulation
+def test_simulate_inductive_load():
+    # -- Setup
+    package_output_dir = PARENT_DIR / 'output' / 'Inductive'
+    package_output_dir.mkdir(parents=True, exist_ok=True)
+    sys_params = SystemParameters(MICROGRID_PARAMS)
+
+    # -- Act
+    inductive = Inductive_load(sys_params)
+    inductive.build_from_template(package_output_dir)
+
+    runner = ModelicaRunner()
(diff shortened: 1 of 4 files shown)


tests/GMT_Lib/test_gmt_lib.py
```diff
@@ -313,6 +313,31 @@ def test_build_inductive_load():
     # Did the mofile get created?
     assert linecount(package_output_dir / 'Inductive0.mo') > 20
 
+
+@pytest.mark.skip(reason="OMC failure: assertion has been violated during initialization")
```
nllong, line 317: Haha, yup

nllong (approved):
Looks great!

PR #562, opened 2023-06-22
Enable numberOfIntervals run option

nllong:
We were only using `stepSize` argument when running openmodelica,

* Add `number_of_intervals` as a simulation option. The simulation will favor `stepSize` if both are given.

* unit tests should suffice

n/a

### DIFF
--- a/geojson_modelica_translator/modelica/modelica_runner.py
+++ b/geojson_modelica_translator/modelica/modelica_runner.py
@@ -102,19 +102,26 @@ class ModelicaRunner(object):
         the generation of the OpenModelica scripts to load and compile/run
         the simulation.
 
-        If passing the start, stop, and step times, then at least start and stop must be supplied to
-        update the simuation to use these values instead of the defaults.
+        Must pass start_time, stop_time, and either step_size or number_of_intervals.
 
         Args:
             run_path (Path): Path where the model will be run, this is where the files will be copied.
             filename (str): name of the file that will be loaded (e.g., BouncingBall.mo, package.mo)
             model_name (str): name of the model to run (e.g., BouncingBall, Districts.DistrictModel)
+            **kwargs: Arbitrary keyword arguments.
+                project_in_library (bool): If True, then the file_to_load is in the library, otherwise it is a file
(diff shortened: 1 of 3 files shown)


geojson_modelica_translator/modelica/modelica_runner.py
```diff
@@ -102,19 +102,26 @@ def _copy_over_docker_resources(self, run_path: Path, filename: Union[str, Path,
         the generation of the OpenModelica scripts to load and compile/run
         the simulation.
 
-        If passing the start, stop, and step times, then at least start and stop must be supplied to
-        update the simuation to use these values instead of the defaults.
+        Must pass start_time, stop_time, and either step_size or number_of_intervals.
```
vtnate, line 105: We're requiring all these now?

geojson_modelica_translator/modelica/modelica_runner.py
```diff
@@ -102,19 +102,26 @@ def _copy_over_docker_resources(self, run_path: Path, filename: Union[str, Path,
...
 
         Args:
             run_path (Path): Path where the model will be run, this is where the files will be copied.
             filename (str): name of the file that will be loaded (e.g., BouncingBall.mo, package.mo)
             model_name (str): name of the model to run (e.g., BouncingBall, Districts.DistrictModel)
+            **kwargs: Arbitrary keyword arguments.
```
vtnate, line 111: Why use kwargs instead of explicitly setting them as parameters of this method?

vtnate (approved):
Jinja for life!! 🤣 Couple questions about kwargs for my own education. This looks good.

geojson_modelica_translator/modelica/modelica_runner.py
nllong: i waffled on this too. I do allow none of these args to be passed and OMC will just run with the defaults.

geojson_modelica_translator/modelica/modelica_runner.py
nllong: not requiring.. we can pass none

PR #563, opened 2023-06-22
Stop and remove containers if user cancels a simulation

### LINKED ISSUE #384: GMT modelica runner should kill docker container
if user runs a model then ctrl-c's it, the docker container should be killed

vtnate:
When running tests developers often `ctrl-c` to cancel a test midway through a long-running simulation. The container is then left alive unnecessarily which is a resource drain.
Catch KeyboardInterrupt while simulating and `kill` all containers from the `gmt-om-runner` image
- Start a simulation test
    - For instance `tests/GMT_Lib/test_gmt_lib.py::test_simulate_cooling_plant`
- `ctrl-c` once the simulation begins
- `docker ps -a` to check that containers from the image are removed
Resolves #384

### DIFF
--- a/geojson_modelica_translator/modelica/lib/runner/om_docker.sh
+++ b/geojson_modelica_translator/modelica/lib/runner/om_docker.sh
@@ -6,19 +6,6 @@
 DOCKER_USERNAME=nrel
 IMG_NAME=gmt-om-runner
 
-# Catch signals to kill the container if it is interrupted
-# https://www.shellscript.sh/trap.html
-# Addresses https://github.com/urbanopt/geojson-modelica-translator/issues/384
-trap cleanup 1 2 3 6
-
-cleanup()
-{
-  echo "Caught Signal ... cleaning up."
-  rm -rf /tmp/temp_*.$$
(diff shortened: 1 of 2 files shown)


vtnate:
@nllong I have a couple questions:
1. Is this functionality even necessary/valuable?
2. The `rm` compound command works as expected when I manually use it in the terminal, but in the code as written it fails with `unknown shorthand flag: 'a' in -a` when I use it. Am I not understanding how `subprocess.run()` works? I've confirmed that the split transforms to a list as expected.

nllong:
> @nllong I have a couple questions:
> 
> 1. Is this functionality even necessary/valuable?
> 2. The `rm` compound command works as expected when I manually use it in the terminal, but in the code as written it fails with `unknown shorthand flag: 'a' in -a` when I use it. Am I not understanding how `subprocess.run()` works? I've confirmed that the split transforms to a list as expected.

I think I just experienced this myself, so yes to 1.
2. -- no idea!

nllong (approved):
fancy! love it...

PR #566, opened 2023-06-25
Breakout package parser class

nllong:
We had two classes in the input_parser file. That's not the best practice.

* Move PackageParser to its own file
* Add in spelling dictionary for use with VSCode's Spell Checker's extension
* Add typing info to PackageParser and InputParser

unit tests should suffice

n/a

### DIFF
--- a/geojson_modelica_translator/modelica/input_parser.py
+++ b/geojson_modelica_translator/modelica/input_parser.py
@@ -64,7 +64,7 @@ class InputParser(object):
         obj_data = ""
         connect_data = ""
         with open(self.modelica_filename, "r") as f:
-            for _, line in enumerate(f.readlines()):
+            for line in f.readlines():
                 if line == "\n":
                     # Skip blank lines (for now?
                     continue
--- a/geojson_modelica_translator/modelica/input_parser.py
+++ b/geojson_modelica_translator/modelica/input_parser.py
@@ -3,134 +3,27 @@
 
 import os
 from pathlib import Path
-from typing import Union
-
-from jinja2 import Environment, FileSystemLoader
-
-from geojson_modelica_translator.jinja_filters import ALL_CUSTOM_FILTERS
-
-
-class PackageParser(object):
-    """
-    Class to read and modify the package.mo and the package.order file
-    """
-
-    def __init__(self, path=None):
-        """
-        Create an instance to manage the package.mo/order file. If no path is provided then the user
-        must add in their own package and order data. Or the user can load from the new_from_template
-        class method.
-
-        :param path: string, path to where the package.mo and package.order reside.
-        """
-        self.path = path
-        self.order_data = None  # This is stored as a string for now.
-        self.package_data = None
-        self.load()
-
-        self.template_env = Environment(
-            loader=FileSystemLoader(
-                searchpath=os.path.join(
-                    os.path.dirname(os.path.abspath(__file__)), "templates"
-                )
-            )
-        )
-        self.template_env.filters.update(ALL_CUSTOM_FILTERS)
-
-    @classmethod
-    def new_from_template(cls, path: Union[str, Path], name: str, order: list[str], within: str = None) -> "PackageParser":
-        """Create new package data based on the package.mo template. If within is not specified, then it is
-        assumed that this is a top level package and will load from the package_base template.
-
-        Args:
-            path (str): the path where the resulting package file and order will be saved to.
-            name (str): the name of the model
-            order (list[str]): ordered list of which models will be loaded (saved to package.order)
-            within (str, optional): name where this package is within.. Defaults to None.
-
-        Returns:
-            PackageParser: object of the package parser
-        """
-        klass = PackageParser(path)
-        if within:
-            template = klass.template_env.get_template("package.mot")
-        else:
-            template = klass.template_env.get_template("package_base.mot")
-
-        klass.package_data = template.render(within=within, name=name, order=order)
-        klass.order_data = "\n".join(order)
-        return klass
-
-    def load(self):
-        """
-        Load the package.mo and package.mo data from the member variable path
-        """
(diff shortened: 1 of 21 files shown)


vtnate:
@nllong 2 tests failed. For me they passed locally, then when I deleted my old docker image (1.21) and the test pulled the new image you pushed (1.20) these tests failed locally. I'm guessing something needs to be changed in the test setup to work with 1.20. Thanks OM!
- tests/GMT_Lib/test_gmt_lib.py::test_simulate_cooling_plant
- tests/model_connectors/test_teaser.py::TeaserModelConnectorSingleBuildingTest::test_simulate_teaser_single

The message in stdout.log is: `Simulation execution failed for model` followed by this output for the Teaser test (the cooling plant only complained about dividing-by-zero at time zero, though its initialization finished successfully as well) :

> LOG_SUCCESS       | info    | The initialization finished successfully with 3 homotopy steps.
assert            | info    | The following assertion has been violated at time 3600.000000
|                 | |       | TeaserLoad_d4487294.ports_aChiWat[1].h_outflow >= -10000000000.0 and TeaserLoad_d4487294.ports_aChiWat[1].h_outflow <= 10000000000.0
assert            | info    | Variable violating min/max constraint: -10000000000.0 <= TeaserLoad_d4487294.ports_aChiWat[1].h_outflow <= 10000000000.0, has value: nan
assert            | info    | The following assertion has been violated at time 3600.000000
|                 | |       | TeaserLoad_d4487294.ports_aHeaWat[1].h_outflow >= -10000000000.0 and TeaserLoad_d4487294.ports_aHeaWat[1].h_outflow <= 10000000000.0
assert            | info    | Variable violating min/max constraint: -10000000000.0 <= TeaserLoad_d4487294.ports_aHeaWat[1].h_outflow <= 10000000000.0, has value: nan
assert            | debug   | division by zero at time 3600.000000360138, (a=0) / (b=0), where divisor b expression is: 0.0

Finally there is this warning, which, while I don't think it's our current issue, may cause some headaches in the future: `Warning: The model contains alias variables with redundant start and/or conflicting nominal values. It is recommended to resolve the conflicts, because otherwise the system could be hard to solve. To print the conflicting alias sets and the chosen candidates please use -d=aliasConflicts.`

vtnate:
> Maybe merge the 1.20 downgrade PR first and then merge develop into here?

I rebuilt the container locally with 1.21 and both of these troublesome tests passed.

nllong:
> > Maybe merge the 1.20 downgrade PR first and then merge develop into here?
> 
> I rebuilt the container locally with 1.21 and both of these troublesome tests passed.

barf. I might just mark those as "skip" until we have time to look into them more?

vtnate:
> 

I pushed a commit that skips those 2 tests.

geojson_modelica_translator/modelica/package_parser.py
```diff
@@ -0,0 +1,124 @@
...
+    def __init__(self, path: Union[str, Path] = None):
+        """Create an instance to manage the package.mo/order file. If no path is provided then the user
+        must add in their own package and order data. Or the user can load from the new_from_template
+        class method.
+
+        Args:
+            path (Union[str, Path], optional): path to where the package.mo and package.order reside.
+                                               Defaults to None.
+        """
+        self.path: Union[str, Path, None] = path
```
vtnate, line 26: Is this redundant because the init parameter has the typing, or is it useful because it's a separate variable? I'm legit asking, not tryna being a PITA.

geojson_modelica_translator/modelica/package_parser.py
vtnate, line 1: All uses of the `os` library in this file can be replaced with `pathlib`. I'm happy to make a commit doing that.

tests/GMT_Lib/test_gmt_lib.py
```diff
@@ -159,6 +119,47 @@ def test_simulate_cooling_plant():
     assert success is True
 
 
+def test_build_community_pv():
```
vtnate, line 122: I moved these tests to a different place in the same file because they weren't with their friends. Is more better now.

vtnate (approved):
Organization? 😱 🙌  I have a couple questions, and in the new package_parser.py we can do everything with pathlib, no need for os.path. I can make that commit, if you are busy. 

I added a commit that skips the 2 tests that are a problem on 1.20. OM really is a finicky beast. Overall this is great, none of my questions stop this from doing what it's supposed to do.

geojson_modelica_translator/modelica/input_parser.py
```diff
@@ -3,134 +3,27 @@
...
-
-from geojson_modelica_translator.jinja_filters import ALL_CUSTOM_FILTERS
 
+class InputParser(object):
+    """Class to read in Modelica files (.mo) and provide basic operations.
 
-class PackageParser(object):
-    """
-    Class to read and modify the package.mo and the package.order file
+    This class is not recommended to be used and ModelicaBuilder should be used
```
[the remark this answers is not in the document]
nllong: I think this will probably stay around for a bit, but the input_parser should def be deprecated and use the modelica builder. I'm working on that right now.

geojson_modelica_translator/modelica/package_parser.py
nllong: `mypy` was throwing errors when passing `self.path` to methods--saying it didn't know the type. So this forces the types. I think I had some `Any`'s in there, which was basically me saying I don't want to deal with it yet.

geojson_modelica_translator/modelica/package_parser.py
nllong: sure, but let's wait until I finish my modelica package parser.

vtnate: Ugh mypy is frustrating 😆. I came around to your way of thinking, and undid my shenanigans. It's happy locally now at least.

PR #568, opened 2023-06-28
Improve gfunction path

vtnate:
The code to read the ghe_dir worked, but was brittle - it only worked if the user was in the GMT root.
Borefield.py now checks if the path supplied by the user in the sys-param file is absolute, and if so, uses it. If it's not absolute, it assumes the path is relative to the sys-param file. The schema entry was updated to say exactly that.
Try supplying either an absolute path or a path relative to the sys-param file (`./`) and running tests/model_connectors/test_borefield.py::test_build_district_system. Try supplying bogus paths to confirm the graceful error message appears.

### DIFF
--- a/geojson_modelica_translator/model_connectors/plants/borefield.py
+++ b/geojson_modelica_translator/model_connectors/plants/borefield.py
@@ -97,7 +97,11 @@ class Borefield(PlantBase):
         }
 
         # process g-function file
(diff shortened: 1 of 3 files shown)


JingWang-CUB (commented):
This looks good to me!

JingWang-CUB (approved):

PR #573, opened 2023-07-31
test a new poetry gha

vtnate:
Poetry often fails multiple times to install dependencies, especially (but I think not exclusively, IIRC) on Windows CI runs. I wanted to see if using a different GitHub Action to run Poetry would make a difference. I think it fails less often now, but definitely not zero failures 😢 .

The solution thus far is to re-run the failed jobs until they magically pass. I don't know why it works sometimes but not others.
Uses a different GitHub Action source for installing Poetry
Run CI and hope it passes all checks on the first try.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,6 @@ jobs:
         run: |
           poetry --version
           poetry install
-          poetry update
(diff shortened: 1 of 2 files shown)


nllong (approved):
sweet

PR #575, opened 2023-08-08
Add new `ModelicaProject` class

nllong:
In the GMT and in ModelicaBuilder, there is no concept of a class that knows of all the files in a Modelica project. In this case, a project is a root `package.mo` that defines the project (but has no `within` block) since it is the root.

This type of method is needed to programmatically create clones of projects that can then change values parametrically to create an analysis.

* New class for loading in all the files related to a Modelica project.
* The `teaser_single` project was copied as the exemplar test project
* On project.save_as(...), the project rename attempts to rename all of the objects that contain the `root` model defined. This is an important note because it is likely that other projects do not use the same conventions and should use this method with caution as it probably will miss some renames.
 
* run the new unit tests for `test_modelica_project.py` and run the resulting `output/test...` files in Dymola

n/a


vtnate:
@nllong [This error](https://github.com/urbanopt/geojson-modelica-translator/actions/runs/5804365015/job/15733827052?pr=575#step:9:435) looks like the test filepath needs to be Windows-aware.

vtnate (approved):
I'm not yet up to speed on modelica-builder, but the new tests pass and the resulting model simulates in OM 1.21 on Ubuntu 22.04.

I made an inconsequential docstring edit, and also updated the poetry lock file.

LGTM!

PR #576, opened 2023-08-09
remove ground loads from system parameter

tanushree04:

Ground loads are computed and stored in memory in ThermalNetwork. They are then passed on to GHE Designer. They do not need to be stored in the system parameter file.

- Remove Ground loads from system parameter > ghe schema and example files

`poetry run py.test tests/system_parameters/tes_systemeters.py`


mitchute (approved):

vtnate (approved):
I love simplifying things!

PR #579, opened 2023-08-29
initialize empty variable before potential use

vtnate:
When running the GMT via the CLI on a Windows computer we found a situation where we could reference a variable before assigning it.
Initialize the variable as an empty string first.


nllong (approved):
dunzo.

PR #584, opened 2023-09-16
Allow skipping of specific files when cloning a modelica project

nllong:
There are certain files that should always be skipped when reading/cloning a modelica project. Maybe this needs to be configurable at some point, but for now it is hardcoded.

* adds list of files to always skip
* moves logtimes to a debug call.

CI should suffice, nothing crazy in here.

n/a


vtnate (approved):
Simplify, simplify.

PR #586, opened 2023-09-25
Microgrid heating

vtnate:
Microgrids can include heating as well as electrical components. This adds a simple boiler to the microgrid project.
- New simple boiler template
- Code to render the template using sys-params
- A test to confirm that the model is generated and simulates successfully
- Use more realistic values for boiler massflow rates in test sys-param files
CI is sufficient, IMHO


geojson_modelica_translator/modelica/GMT_Lib/DHC/Components/Plants/Heating/BoilerPolynomial.mot
```diff
@@ -0,0 +1,43 @@
...
+  Documentation(info="<html>
+    <p>This model validates the hot water boiler template model implemented in
+<a href=\"Buildings.Fluid.Boilers.Examples.BoilerPolynomial\">
+Buildings.Fluid.Boilers.Examples.BoilerPolynomial.mot</a>.
+</p>
+</html>",
+        revisions="<html>
+<ul>
+<li>
+September 10, 2023 by Saranya Anbarasu:<br/>
```
nllong, line 34: Add an <li> with date and your name that you "templated for the GMT"

geojson_modelica_translator/modelica/GMT_Lib/DHC/Components/Plants/Heating/BoilerPolynomial.mot
```diff
@@ -0,0 +1,43 @@
+within;
+model BoilerPolynomial
+"Hot water boiler template model with both steady and dynamic state implementation for GMT level 1 testing"
+extends Modelica.Icons.Example;
+
+parameter Modelica.Units.SI.Power Q_flow_nominal_GMT = {{ data["heat_flow_nominal"] }} "Nominal power";
+parameter Modelica.Units.SI.MassFlowRate m_flow_nominal_GMT = {{ data["mass_hhw_flow_nominal"] }} "Nominal mass flow rate,
```
nllong, line 7: not sure I like the `_GMT` appended name. Why not just m_flow_nominal...

nllong (changes requested):
sweet, if the tests pass then I think we are good on the modelica side as it looks like the model is heavily wrapped within modelica.

geojson_modelica_translator/modelica/GMT_Lib/DHC/Components/Plants/Heating/BoilerPolynomial.mot
vtnate: @nllong We need the name to be different in some way, due to the underlying MBL model. The original boiler model from MBL includes the same parameters declaration. Since the parameters are not replaceable from the original MBL boiler model, we cannot redeclare the parameters to overwrite the value. Happy to have any other string appended, or difference in name if there is something you would prefer.

geojson_modelica_translator/system_parameters/schema.json
```diff
@@ -842,19 +842,19 @@
       "type": "object",
       "properties": {
         "heat_flow_nominal": {
-          "description": "[unused] Nominal district heating load. (W)",
+          "description": "Nominal district heating load. (W)",
```
nllong, line 845: cool

nllong (approved):
lets gooo

PR #589, opened 2023-09-27
Add a quick fix for reading gfunction.csv from ghe_id subfolder

JingWang-CUB:
GTO URBANopt project

Add a quick fix to avoid failing in reading gfunction from URBANopt GHE output folders.

poetry run pytest -m 'not compilation and not simulation' tests/model_connectors/test_district_single_ghe.py

None


vtnate (approved):
LGTM! I love that we could simplify the tests and only have a single location for the gfunction.csv.

PR #590, opened 2023-09-28
Change how ModelicaPaths are built so they also work on Windows computers

vtnate:
We weren't modifying paths correctly for use inside Modelica on Windows machines. Fixing this gets us one step closer to supporting Windows in the GMT.

Note the backslash and unknown character as attempted separators in this Dymola error message:
`Not possible to open file "D:/GitHub/geojson-modelica-translator/tests/model_connectors/output/district_single_ghe/Loads/Resources\Databcdefghijklmnopqrstuvwx/time_series_large_office_2013_5B.mos" for reading:
Invalid argument`
Changes the output of the ModelicaPath class to always use `/` in paths rather than being platform-aware. Modelica only uses forward-slash, even on Windows. This was not an issue when simulating on Linux because the platform-aware method produced the `/` as Modelica wanted.
- Run any `not compilation and not simulation` test
- Open the model in Dymola/OpenModelica on your Windows computer
- Confirm that it compiles/translates appropriately
- Confirm that the model simulates as expected


JingWang-CUB (approved):
Everything looks good to me. I'm happy that this is fixed!

PR #591, opened 2023-09-29
Move modelica methods from GMT to modelica-builder

nllong:
There was a growing list of Modelica-based methods that were making their way into the GMT. These classes probably belong best in another repo.

* Move the Modelica methods to Modelica-Builder: ModelicaProject, PackageParser, and ModelicaMOSFile.

* There is no functionality change, just library imports. 

* Before merge, we need to release a new version of ModelicaBuilder and update the pyproject file.

n/a


vtnate (approved):
Separation of church and state ❤️

PR #601, opened 2023-11-21
Refactoring and enhancing to support multiple GHEs in a single district

vtnate:
The MVP currently available only supports a single GHE in the sys-param file. In reality multiple GHEs are common.
- Refactor system_parameters.get_param_by_building_id into .get_param_by_id

Additional work is required to loop through GHEs and build multi-GHE models appropriately. The strategy for that is being developed (modelica arrays or GMT-style individual models?).
CI should be sufficient. The important files are system_parameters.py & test_system_parameters.py.

Other changes:
- Refactoring to use the renamed method
- Ensuring there are unique ghe_ids in a single sys-param file and the same ghe_ids in different test files so they work together.


geojson_modelica_translator/system_parameters/system_parameters.py
```diff
@@ -149,22 +149,24 @@ def get_param(self, jsonpath, data=None, default=None, impute_default=True):
...
         :return: variant, the value from the data
         """
 
         for b in self.param_template.get("buildings", []):
-            if b.get("geojson_id", None) == building_id:
+            if b.get("geojson_id", None) == id:
                 return self.get_param(jsonpath, data=b)
+        for ghe in self.param_template.get("district_system")["fifth_generation"]["ghe_parameters"]["ghe_specific_params"]:
+            if ghe.get("ghe_id", None) == id:
+                return self.get_param(jsonpath, data=ghe)
```
vtnate, line 167: Additional blocks can be added here to support future lists of items in the sys-param file

nllong (approved):
Looks good. So this will require that the IDs are unique across the entire GeoJSON file, correct? It should, since it is an ID, but wanted to make sure I understood the implication. Thanks!

Issue #598, opened 2023-11-06
Create a Batch File Version of the om_docker.sh That Can be Run on Windows


chriswmackey:
Currently, when trying to run the GMT on Windows, we aren't able to call Docker since a shell script is being used to do this and Windows cannot run .sh files natively (typically, this is done with a batch script in Windows instead of shell). So, to get this to work on Windows we need a `om_docker.bat` version of [this om_docker.sh](https://github.com/urbanopt/geojson-modelica-translator/blob/develop/geojson_modelica_translator/modelica/lib/runner/om_docker.sh).

Or it may be better practice to move some of the logic in that shell script to be run using Python in the GMT package and just use Python `subprocess` to call ` docker run`, which seems to be the main part of the shell script that is calling docker:

https://github.com/urbanopt/geojson-modelica-translator/blob/develop/geojson_modelica_translator/modelica/lib/runner/om_docker.sh#L107

Whatever the case, calling a .sh file using Python `subprocess` as what happens on this line here will definitely fail in Windows:
https://github.com/urbanopt/geojson-modelica-translator/blob/develop/geojson_modelica_translator/modelica/modelica_runner.py#L176

So this either needs to change to call a batch file or another executable that will be recognizable on both operating systems (like `docker run`).
I'm happy to look into this and come up with something that will work on Windows.

vtnate:
@chriswmackey Yes, let's discuss. If we can do it all in Python and eliminate the shell script altogether I'd be delighted. The need for the shell script may have passed when we got pushed off JModelica.

PR #606, opened 2023-12-15
fix(modelica_runner): Replace shell script with call directly to docker

chriswmackey:
Resolves https://github.com/urbanopt/geojson-modelica-translator/issues/598

This PR essentially replaces the `om_docker.sh` script with a direct call to `docker run`, thereby enabling the docker simulation to work on Windows (where shell scripts are not executable) just as it does on unix-based systems.

Using poetry I have verified that the implementation in this PR succeeds for practically all tests on my machine. 119 tests passed on my machine and the 9 that seem to have failed all appear to be Dymola, which would make sense since I do not have Dymola installed.

image

The only thing that needs to be done for someone to be able to run all of the tests on their end is that a new docker image generated from the updated Dockerfile has to be pushed to nrel/gmt-om-runner on dockerhub. This will ensure that there's an image on dockerhub that has the MBL installed.

Alternatively, if you want to run some tests before the the new image is pushed to Dockerhub, just build the image yourself by running `docker build` on the Dockerfile in this PR. Then replace [this line of code in the GMT](https://github.com/chriswmackey/geojson-modelica-translator/blob/develop/geojson_modelica_translator/modelica/modelica_runner.py#L165) with the ID of your locally-built image. This is how I verified that all of the tests are passing for me.

CC: @vtnate


vtnate:
Holy macaroons @chriswmackey this is glorious! 💯 

- Can you fix the merge conflicts? Github appears to be telling me I am not allowed to.
- When you run your tests locally, you can deselect the Dymola tests by adding `-m "not dymola"` to the end of the pytest command.
- The other test failures appear to be looking for Windows filepaths, so I bet you have somewhere that is being platform-aware or platform-unaware causing that problem

This is very exciting!

vtnate:
I merged a PR into develop since you branched, which makes a bunch of tests better. Can you merge the latest gmt-develop into your branch too?

chriswmackey:
Hey @vtnate . I think I successfully rebased everything such that there are no merge conflicts. The only thing I was not sure about is that one of the other commits that had been merged in the last few days added this cmake dependency into the docker image:

https://github.com/urbanopt/geojson-modelica-translator/pull/606/files#diff-6b42c56f0c9822944bb23e2929be1c8e09fb50f6079fec8171e9915944957a06R8

I have removed it in my PR since it wasn't necessary for all of the tests I was running but just let you know if you need me to add it back.

vtnate:
Fascinating! Yeah, our runs were breaking without cmake, but if this new way of building it doesn't need it, great!

One final request: can you update the docker base image to pull from `v1.22.1` instead of `v1.22.0`?

I wonder why our CI isn't picking this up and running? Maybe we should merge it into a branch first, and then I can try merging that branch into develop?

chriswmackey:
Actually, I realized that all of those tests I had run were using the old OM docker image as a base. So I just put back the cmake since I think you may be right that it's needed to work with this version of OM. I also changed the base image to pull from v1.22.1.

I'm re-running the tests locally on my machine now.

Just let me know if you want me to send this PR elsewhere into a different branch.

vtnate:
I just made a `new-docker` branch, can you set that as the base branch for this PR?

vtnate:
Never mind, I can change the base branch

geojson_modelica_translator/modelica/lib/runner/Dockerfile
```diff
@@ -15,5 +15,8 @@ RUN echo \
   omc /tmp/installMSL.mos && \
   rm /tmp/installMSL.mos
 
-# Put the om.py CLI into the container
-COPY ./om.py /mnt/lib/om.py
+# Install MBL 9.1.0
+RUN echo \
+  "updatePackageIndex(); getErrorString(); installPackage(Buildings, \"9.1.0\", exactMatch=false); getErrorString();" >> /tmp/installMBL.mos && \
```
vtnate, line 20: 🙌

geojson_modelica_translator/modelica/lib/runner/om_docker.sh
vtnate, line 1: Bye Felicia!

vtnate (approved):

PR #607, opened 2023-12-15
Replace shell script with call directly to docker

### LINKED ISSUE #598: Create a Batch File Version of the om_docker.sh That Can be Run on Windows

vtnate:
Now that JModelica isn't a thing, and we aren't using a branch of the MBL, we can have a cleaner simulation workflow. As a "side benefit" we also support Windows computers now!

Huge thanks to @chriswmackey for making this happen
- Remove om_docker.sh and om.py
    - These spun up a docker container, connected it to the user's local MBL, and provided a cli inside the container to run JModelica
    - We used the local MBL because we had to use a branch at the time. Now that we use released versions of the MBL, there is no need for that step
- Installs MBL right into the docker container
- Runs the container and sends commands straight to OM from modelica_runner.py
    - This is suuuuuuuch a nice improvement!
- Build the container locally
- Update [this line](https://github.com/urbanopt/geojson-modelica-translator/blob/new-docker/geojson_modelica_translator/modelica/modelica_runner.py#L154) to point to your newly-built container
- Run pytest with all the (non-dymola) simulations, especially if you have a Windows computer!
Resolves #598


.github/workflows/ci.yml
```diff
@@ -59,61 +59,52 @@ jobs:
...
-          find $GITHUB_WORKSPACE -type f -name 'stdout.log' -print | while read filename; do
-            echo "============================================ stdout.log ========================================="
-            echo "$filename"
-            cat "$filename"
-          done
+        if: matrix.test_env == 'python' && matrix.os == 'ubuntu-latest'
+        run: poetry run pytest tests/model_connectors/test_district_single_ghe.py --doctest-modules -v -m 'not dymola' --cov-report term-missing --cov . ./tests
+      # - name: Run Pytest (no compilation or simulation)
+      #   if: matrix.test_env == 'python' && matrix.os == 'windows-latest'
+      #   run: poetry run pytest --doctest-modules -v -m 'not simulation and not compilation and not dymola' ./tests
```
nllong, line 81: do we need to uncomment all of these?

geojson_modelica_translator/modelica/modelica_runner.py
```diff
@@ -166,14 +150,14 @@ def _subprocess_call_to_docker(self, run_path: Union[str, Path], action: str) ->
...
-            # get the relative difference between the file to run and the path which everything is running in.
-            # make sure to simulate at a directory above the project directory!
-
-            # Use slashes for the location of the model to run. We can make these periods `.replace(os.sep, '.')`
-            # but must strip off the .mo extension on the model to run
-            # run_model = Path(file_to_run).relative_to(run_path)
-            exec_call = ['./om_docker.sh', action]
+            # create the command to call the open modelica compiler inside the docker image
+            exec_call = ['docker', 'run', '-v', f'{run_path}:/mnt/shared/{model_name}', f'{image_name}',
+                         '/bin/bash', '-c', f"cd mnt/shared/{model_name} && omc {mo_script}.mos"]
```
nllong, line 159: noice!

nllong (approved):

Issue #609, opened 2024-01-05
Enable user configuration of loop pump nominal head pressure


vtnate:
### Expected Behavior

Users should be able to set the head pressure of the district pump

- It should be added to the schema, probably for both 4th and 5th gen systems. So, [here](https://github.com/urbanopt/geojson-modelica-translator/blob/develop/geojson_modelica_translator/system_parameters/schema.json#L1193) and [here](https://github.com/urbanopt/geojson-modelica-translator/blob/develop/geojson_modelica_translator/system_parameters/schema.json#L1224).
- Then it should be added into a test file, like [here](https://github.com/urbanopt/geojson-modelica-translator/blob/develop/tests/data_shared/system_params_des_5g.json#L129), and a test should be made (or an assertion added to an existing test) that checks that it gets read and used properly.

### Actual Behavior

It's hard-coded [here](https://github.com/urbanopt/geojson-modelica-translator/blob/mblv10/geojson_modelica_translator/model_connectors/networks/templates/NetworkDistributionPump_Instance.mopt#L25). A larger district would be too much for this pump setting, causing a modelica simulation failure.

### Version Information

Modelica Buildings Library: `10.0.0`
Modelica Standard Library: `4.0.0`
Python: `3.11.7`

JingWang-CUB:
Hi @tanushree04 , can you please add the "pump_design_head" parameter to "central_pump_parameters" under the 4th and 5th gen system schema? An example can be found here: https://github.com/urbanopt/geojson-modelica-translator/pull/610/files#. Let me know if you have any questions.

PR #610, opened 2024-01-09
Add district nominal pump head to system parameters file

JingWang-CUB:
MBLv10 has changes to their pump model which sets an upper bound to the actual pump head based on the nominal head. This has caused the 5G district models to fail if the nominal pump head is not set correctly.

This PR fixes the above issue by allowing the users to change the nominal district pump head from the system parameters file.

poetry run pytest -m 'not dymola' tests/model_connectors/test_district_single_ghe.py

#609

### DIFF
--- a/tests/model_connectors/data/system_params_ghe.json
+++ b/tests/model_connectors/data/system_params_ghe.json
@@ -40,6 +40,9 @@
   ],
   "district_system": {
     "fifth_generation": {
+      "central_pump_parameters": {
+        "pump_design_head": 60000
+      },
       "ghe_parameters": {
         "version": "1.0",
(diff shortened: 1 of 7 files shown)


tests/model_connectors/data/system_params_ghe.json
```diff
@@ -40,6 +40,9 @@
   ],
   "district_system": {
     "fifth_generation": {
+      "central_pump_parameters": {
+        "pump_design_head": 60000
+      },
```
tanushree04, line 45: @JingWang-CUB what are the units for this?

JingWang-CUB: Pa

vtnate (approved):
🌸

PR #617, opened 2024-01-17
Update instructions for releasing documentation publicly

vtnate:
The way we release the library in general and the documentation for this library has been updated. The instructions now match reality.
- Remove references to JModelica that were left over from the bad old days
- Make the documentation release instructions match our current practices


nllong (approved):
🤘

PR #618, opened 2024-01-22
Rename new modelica models to clarify MBL version

### LINKED ISSUE #612: Rename new models for clarity
### Expected Behavior
Model name should be more descriptive of what it applies to

### Actual Behavior
- [This file](https://github.com/urbanopt/geojson-modelica-translator/blob/develop/geojson_modelica_translator/model_connectors/plants/templates/Boiler_TParallel_new.mo) (and model) has a`_new` name that will get confusing in the future. We should rename them. Perhaps something like `_v10` to indicate what version of MBL it is compatible with?
- The same should be done with [this one](https://github.com/urbanopt/geojson-modelica-translator/blob/develop/geojson_modelica_translator/model_connectors/plants/templates/PartialPlantParallel_v1new.mo).

### Version Information

Modelica Buildings Library: `10.0.0`
Modelica Standard Library: `4.0.0`
TEASER: `X.Y.Z`
Python: `3.11.7`

vtnate:
We had to make some new models to support MBLv10. They were just named `..._new` which could get confusing once we forgot when/why these were created.
- Renames 2 models to clarify that they work with/are intended for MBLv10
- Update Docker readme to reference current version of our Dockerfile
    - This is not substantive. I happened to notice it when rebuilding the image locally

@nllong What naming convention should we use?
CI is sufficient
Resolves #612

### DIFF
--- a/geojson_modelica_translator/model_connectors/plants/chp.py
+++ b/geojson_modelica_translator/model_connectors/plants/chp.py
@@ -25,11 +25,11 @@ class HeatingPlantWithOptionalCHP(PlantBase):
             self.id = 'heaPla' + simple_uuid()
 
         self.required_mo_files.append(Path(self.template_dir) / 'Boiler_TParallel.mo')
-        self.required_mo_files.append(Path(self.template_dir) / 'Boiler_TParallel_new.mo')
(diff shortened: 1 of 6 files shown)


nllong (approved):
sweet, I like it!

PR #622, opened 2024-03-05
Enable more detail when specifying district types

vtnate:
BREAKING CHANGE - CLI

Our existing technique for specifying which type of district we're modeling (4G, 5G, 5G with GHE, Steam) was not very simple or extensible. This should make it both more clear and better able to handle future additions. @nllong Let's talk about how we name the district types/generations.
- Changes the CLI input from a boolean flag (`--ghe`) to be a string from the user specifying what type of district they are modeling
- Code to support that
- Updates to tests to handle the new CLI command and code changes
CI is sufficient


JingWang-CUB (approved):
Hi Nate, these changes look good to me. It's an exciting change.

nllong (approved):
love it, cleaner for sure!

To you naming questions... we should figure out a table like this. 

https://github.com/urbanopt/geojson-modelica-translator-examples?tab=readme-ov-file#matrix-of-examples

PR #626, opened 2024-03-11
Force a dummy value for SHW in modelica loads even if not present

vtnate:
MBLv10 requires a value in the Service Hot Water column of the loads timeseries, even if it isn't used. Modelica-Builder [has a feature](https://github.com/urbanopt/modelica-builder/blob/develop/modelica_builder/modelica_mos_file.py#L93) that can update values in .mos files, so lets use that.

_The UO SDK populates this column as of v0.11.0 with https://github.com/urbanopt/urbanopt-reporting-gem/pull/143. Load files generated before then may be missing this data._
- Uses the modelica-builder method to set SHW load to 1/10 of the heating load or 5000W, whichever is higher.
- Improve logging a bit
Generate a model using timeseries data that does not have the SHW column populated. The updated file is `modelica_project/Loads/Resources/Data/<buildingname>/<load_timeseries>.mos`

### DIFF
--- a/geojson_modelica_translator/utils.py
+++ b/geojson_modelica_translator/utils.py
@@ -60,12 +60,17 @@ def _add_water_heating_patch(modelica_dir: Path):
     if data_dir.is_dir():
         for bldg_dir in data_dir.iterdir():
             mo_load_file = data_dir / bldg_dir / "modelica.mos"
+            # In case the modelica loads file isn't named modelica.mos:
+            if not mo_load_file.is_file():
+                modelica_loads = list((data_dir / bldg_dir).rglob("*"))
+                if len(modelica_loads) == 1:
+                    mo_load_file = modelica_loads[0]
             if mo_load_file.is_file():
                 fixed_lines, fl_found = [], False
                 with open(mo_load_file) as mlf:
                     for line in mlf:
                         if line == "#Peak water heating load = 0 Watts\n":
-                            logger.debug(f"Adding dummy value for water heating for {bldg_dir}")
+                            logger.debug(f"Adding dummy value for water heating to {mo_load_file}")
                             nl = "#Peak water heating load = 1 Watts\n"
                             fixed_lines.append(nl)
                         elif not fl_found and ";" in line:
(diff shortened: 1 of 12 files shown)


geojson_modelica_translator/utils.py
```diff
@@ -7,7 +7,12 @@
 from pathlib import Path
 from uuid import uuid4
 
-_log = logging.getLogger(__name__)
+logger = logging.getLogger(__name__)
+logging.basicConfig(
```
nllong, line 11: if we are going to specify the format, then we should propagate to all the other files, or setup in a config file.

nllong (changes requested):
I wish we chatted... I wrote a helper method to do this in the 5G stuff. The method is ultimately in Modelica Builder.

https://github.com/urbanopt/geojson-modelica-translator/blob/develop/geojson_modelica_translator/modelica/GMT_Lib/DHC/DHC_5G_waste_heat_GHX_variable.py#L67-L77

```
mos_file = ModelicaMOS(save_filename)
            total_heating_load += mos_file.retrieve_header_variable_value('Peak space heating load', cast_type=float)
            total_cooling_load += mos_file.retrieve_header_variable_value('Peak space cooling load', cast_type=float)
            peak_water = mos_file.retrieve_header_variable_value('Peak water heating load', cast_type=float)
            total_swh_load += peak_water
            if peak_water == 0:
                peak_heat = mos_file.retrieve_header_variable_value('Peak space heating load', cast_type=float)
                peak_swh = max(peak_heat / 10, 5000)

                mos_file.replace_header_variable_value('Peak water heating load', peak_swh)
                mos_file.save()
```

vtnate:
@nllong You mean... have the modelica package do the modelica editing, and keep log messages consistent?

image

nllong:
> @nllong You mean... have the modelica package do the modelica editing, and keep log messages consistent?
> 
> image

I know, it's complicated. haha

nllong (approved):
thanks for using Modelica Builder! love it.

PR #628, opened 2024-03-28
Relax pandas version requirement

vtnate:
The main urbanopt SDK was having troubles with incompatible dependencies. Pandas was the primary culprit. Relaxing this version constraint among all the relevant packages will help us escape dependency hell.

- Python has their general [dependency specification](https://packaging.python.org/en/latest/specifications/version-specifiers/#compatible-release)
- Poetry has [their own specification](https://python-poetry.org/docs/dependency-specification/)
Allows Pandas >=2 <3
CI is sufficient for the mechanics. Happy to have a philosophy discussion if there is a preferred technique to escape dependency hell.
https://github.com/urbanopt/urbanopt-cli/issues/459

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -446,69 +446,69 @@ test = ["pytest (>=6)"]
 
 [[package]]
 name = "filelock"
-version = "3.13.1"
+version = "3.13.3"
 description = "A platform independent file lock."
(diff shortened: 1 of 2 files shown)


nllong (approved):
😍

PR #630, opened 2024-04-06
Expose borehole variables

nllong:
These variables were buried in the borFieDat within a borFie object.

Expose the parameters to make them easier to set.

Tests alone should suffice.


nllong:
> I'm not really sure what's going on here, but I can see how this give modelica-builder access, and, as you say, the tests continue to pass.
> 
> This template (afaik) is only used in tests, not in real running of the GMT. Is this capability already covered in the main borefield templates? https://github.com/urbanopt/geojson-modelica-translator/blob/develop/geojson_modelica_translator/model_connectors/plants/templates/BorefieldOneUTube.mot and https://github.com/urbanopt/geojson-modelica-translator/blob/develop/geojson_modelica_translator/model_connectors/plants/templates/BorefieldTwoUTubes.mot?

I use this template for the 5G systems quite a bit. I would like to move over to a more configurable approach, but that hasn't happened yet, for me. Happy to jump on call.... which I think is needed.

vtnate (approved):

PR #627, opened 2024-03-19
Templates and tests for horizontal piping modeling

JingWang-CUB:
#### Any background context you want to provide?
UO GHP project

#### What does this PR accomplish?
Integrate new templates and tests for geothermal horizontal piping modeling.

1. Added new template `GroundCoupling.mot` and accompanying python file.
2. Included dependent network mo files (`ConnectionSeriesAutosize.mo`, `UndisturbedSoilTemperature.mo`, `UnidirectionalSeries.mo`).
3. Included an external csv file to provide coefficients for calculating soil temperatures (`Soil_temp_coefficients.csv`).
4. Added new coupling `GroundCoupling_Borefield`.
5. Made modifications to tests `test_district_single_ghe` and `test_uo_des` to test the new horizontal piping features.

#### How should this be manually tested?
poetry run pytest tests/model_connectors/test_district_single_ghe.py

PR #632, opened 2024-04-19
Add horizontal pipe properties

tanushree04:
Adds properties for horizontal piping 
#633


geojson_modelica_translator/system_parameters/schema.json
```diff
@@ -26,6 +26,9 @@
             },
             {
               "$ref": "#/definitions/fifth_generation_def"
+            },
+            {
+              "$ref": "#/definitions/horizontal_piping_parameters_def"
```
JingWang-CUB, line 31: The horizontal piping parameters should be added under "fifth_generation_def" following "central_pump_parameters".

geojson_modelica_translator/system_parameters/schema.json
```diff
@@ -1190,6 +1193,49 @@
...
+              "description": "Thickness of pipe insulation, used to compute R. If no insulation, use a virtual insulation layer of soil. Must be greater than zero.",
+              "type": "number"
+            },
+            "insulation_conductivity": {
+              "description": "Heat conductivity of pipe insulation, used to compute R. If no insulation, use a virtual insulation layer of soil. Must be greater than zero.",
+              "type": "number"
+            },
+            "diameter_ratio": {
+              "description": "The ratio of pipe outer diameter to pipe wall thickness",
+              "type": "number"
```
JingWang-CUB, line 1214: @mitchute , should diameter ratio be limited to type "integer"?

JingWang-CUB (changes requested):
Hi @tanushree04 , I left some comments. Otherwise, everything looks good. Should we also update how the default values are propagated, etc.?

geojson_modelica_translator/system_parameters/schema.json
vtnate: @JingWang-CUB Am I correct that this means the horizontal piping work you're doing only applies to 5G systems? Do 4G (and earlier) systems have pipes that these parameters won't apply to?

JingWang-CUB: Because 4G systems will likely have separate heating and cooling pipes so these parameters are not good fits for 4G.

tanushree04:
@JingWang-CUB if we have an example file with horizontal piping parameters we can add those to the tests to test the schema change otherwise this should be wrapped up.

JingWang-CUB:
> @JingWang-CUB if we have an example file with horizontal piping parameters we can add those to the tests to test the schema change otherwise this should be wrapped up.

I have one example file [here](https://github.com/urbanopt/geojson-modelica-translator/pull/627/files#diff-62c760248f29ea0f70c769c71a193c6af73b7d53d657f50b32bf928a691c0c97) on my branch which can be used for testing the schema.

PR #640, opened 2024-06-20
Add default cop values for 5G district heat pump efficiencies

vtnate:
#### Any background context you want to provide?
We should have a user-configurable 5G heat-pump. COP values are important to be able to customize.
#### What does this PR accomplish?
- [x] Adds `wahp` (and sub-items) to sys-param template
-  [x] Add `cop_c` and `cop_h` to schema
#### How should this be manually tested?
Confirm that [this test](https://github.com/urbanopt/geojson-modelica-translator/blob/develop/tests/management/test_uo_des.py#L52) generates a sys-param file (saved into [this folder](https://github.com/urbanopt/geojson-modelica-translator/tree/develop/tests/management/data/sdk_project_scraps/run/baseline_scenario)) that includes `district_system.fifth_generation.wahp`.

PR #642, opened 2024-07-03
Use the correct ETS heat pump COP

vtnate:
Unfortunately #640 was misguided and the COP values we are interested in are contained within the ETS for each building. Fortunately we already have those values populated in the template sys-param file.
Revert everything in #640 except the typo fixes
CI is sufficient


nllong (approved):
very nice...

PR #644, opened 2024-07-11
Support multiple GHEs in a single district

vtnate:
### Any background context you want to provide?
Previously we only supported a single GHE in a thermal loop
### What does this PR accomplish?
- Reads all ghe_specific_params from the sys-param file
- Instantiates a separate borefield object for each GHE
- Couples the district together
### How should this be manually tested?
1. Build a Modelica model using the [multi-ghe test](https://github.com/urbanopt/geojson-modelica-translator/blob/ghe-couplings/tests/model_connectors/test_district_multi_ghe.py#L90)
2. Load the model in Dymola (by hand)
3. Simulate in Dymola
    - OM crashes when trying to open a model that large :(
### Future work:
- Ensure this will simulate natively inside GMT using OM


mitchute (approved):

nllong (approved):
This is hands down one of the most impressive PRs that I have seen. Most excellent work team!

image

PR #648, opened 2024-08-08
Update code & dependencies to address deprecation warnings

vtnate:
Pandas was warning about deprecated code. This addresses all the warnings.
- Use Pandas in a way that will be more stable with future versions
- Use stable release v1.0.1 of Teaser to include the deprecation fixes handled there
    - Adapt our code to their new API interface
CI is sufficient


pyproject.toml
```diff
@@ -42,8 +42,8 @@ modelica-builder = "^0.5.1"
 # modelica-builder = { git = "https://github.com/urbanopt/modelica-builder.git", branch = "mblv10"}
 pandas = "~2"
 requests = "^2.28"
-teaser = "0.7.5"
-#teaser = { git = "https://github.com/urbanopt/TEASER.git", branch = "development"}
+# teaser = "0.7.8"
```
vtnate, line 45: Given the API changes I ran into, I suspect the next version will actually be `v0.8.0` instead of 0.7.8.

nllong: sweet, I can ask about next release date.

nllong (approved):

PR #652, opened 2024-09-03
Use feature ids in load names instead of simple_uuid

### LINKED ISSUE #645: Use building IDs from GeoJSON / Find a way to link Modelica outputs to GeoJSON Feature IDS
We store building IDs in the URBANopt GeoJSON file, while querying outputs from the Modelica Model they ouputs have a UUID associated with the building Feature. We need to find a way to link the output variable UUID to the UO building Feature ID
image

vtnate:
This will simplify access to results of Modelica simulations by naming the Modelica variables with the geojson_id of the load (building). Future work should do the same thing for borefields when modeling GHEs.
Update `self.id` for the loads to use the geojson_id instead of generating a simple_uuid. We retain simple_uuid for objects that aren't being supplied by the geojson file, such as networks, couplings, ETSs, etc.
CI is sufficient
Resolves #645

### DIFF
--- a/geojson_modelica_translator/model_connectors/load_connectors/spawn.py
+++ b/geojson_modelica_translator/model_connectors/load_connectors/spawn.py
@@ -7,7 +7,7 @@ import shutil
 from modelica_builder.package_parser import PackageParser
 
 from geojson_modelica_translator.model_connectors.load_connectors.load_base import LoadBase
-from geojson_modelica_translator.utils import ModelicaPath, convert_c_to_k, simple_uuid
+from geojson_modelica_translator.utils import ModelicaPath, convert_c_to_k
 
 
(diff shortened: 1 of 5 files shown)


tanushree04:
Looks great @vtnate , does this take care of all modelica variables for buildings in the `TimeSerLoa_` format?

vtnate:
> Looks great @vtnate , does this take care of all modelica variables for buildings in the `TimeSerLoa_` format and adds the building feature instead of the UUID?

That's the intent, yes. Can you generate a model from this branch and confirm that it provides the info you need?

tanushree04 (approved):
image

I created a new modelica model and ran the results from my script, the variables are now named as per the feature ID, great work!

PR #653, opened 2024-09-04
Fix docs

vtnate:
The `docs` step of CI was failing because of some versions that have fallen into disrepair.
- Ensures setuptools is available during docs step of CI
- Replaces `m2r2` with `myst_parser` in sphinx config


nllong (approved):
nice fix! Approved... and merged.

PR #654, opened 2024-09-06
Support international weather locations

### LINKED ISSUE #651: Support for international location weather files
The GMT does not currently support running analysis for a location outside of North America, given the current process for downloading weather files. The thermal microgrids ESTCP project, and likely future projects, would like to be able to run analysis thru the GMT for an international location. @JingWang-CUB FYI

vtnate:
We were only supporting USA weather locations until now. This PR enables any location from www.energyplus.net/weather to be used (as of Sep 4, 2024). _**Major caveat**_: This only accepts *.mos weather files, which as of now are only available on the website for USA locations. Once mos weather files for all locations on the website are available, this code is ready to take advantage.
- Handle all locations provided by https://energyplus.net/weather. Requires a weather file in *.mos format to be avaailable from that website, which as of today is limited to USA locations.
- Adds tests for this using the CLI and using the code directly
Provide a valid weather epw filename for any location outside the USA in your geojson file. Observe that the EPW is downloaded correctly and an error is raised if the mos format file isn't found.
Resolves #651

### DIFF
--- a/geojson_modelica_translator/system_parameters/system_parameters.py
+++ b/geojson_modelica_translator/system_parameters/system_parameters.py
@@ -207,12 +207,111 @@ class SystemParameters:
                 "Malformed location, needs underscores of location "
                 "(e.g., USA_NY_Buffalo-Greater.Buffalo.Intl.AP.725280_TMY3.mos)"
             )
-
         # download file from energyplus website
-        weatherfile_url = (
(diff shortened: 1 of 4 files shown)


nllong (approved):
nicely done--looks like we should have handled this from the beginning. Thanks for updating!

PR #656, opened 2024-09-10
Add downloaded weatherfiles to gitignore

vtnate:
#654 added new tests which download new weatherfiles. That test output isn't required to be stored in github.
Gitignore German weather files downloaded via https://github.com/urbanopt/geojson-modelica-translator/blob/develop/tests/management/test_uo_des.py#L73


vtnate (commented):
This was accidentally not included in #654

PR #662, opened 2024-10-07
Improve cli typing & validation

vtnate:
The UO cli was also supplying default values to this cli (in addition to our own default values), which caused collisions in the `create-model` case. Documented in [this issue](https://github.com/urbanopt/urbanopt-cli/issues/480) and resolved in [this PR](https://github.com/urbanopt/urbanopt-cli/pull/486) in the main uo-cli repo.

During diagnosis of that behavior, I learned about additional validation available in [click.Path](https://click.palletsprojects.com/en/8.1.x/api/#click.Path) which I implemented here.
- Sets `path_type` for cli arguments, so they can be used directly in the code (instead of being read as strings, the previous behavior)
- Add a validation step where Click checks that the path leads to a readable file.
CI is sufficient

### DIFF
--- a/management/uo_des.py
+++ b/management/uo_des.py
@@ -22,15 +22,15 @@ def cli():
 @cli.command(short_help="Create sys-param file")
 @click.argument(
     "sys_param_filename",
-    type=click.Path(file_okay=True, dir_okay=False),
+    type=click.Path(file_okay=True, dir_okay=False, path_type=Path, readable=True, resolve_path=True),
 )
 @click.argument(
     "scenario_file",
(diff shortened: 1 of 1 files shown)


management/uo_des.py
```diff
@@ -90,14 +90,14 @@ def build_sys_param(
     """
 
     # Use scenario_file to be consistent with sdk
-    scenario_name = Path(scenario_file).stem
-    scenario_dir = Path(scenario_file).parent / "run" / scenario_name
+    scenario_name = scenario_file.stem
```
nllong, line 93: nice cleanup

nllong (approved):
let's gooooo

PR #668, opened 2024-10-21
Clean up sys-param code that reads parameters

### LINKED ISSUE #495: Refactor `get_param` since there is no longer a default section of the sys-param file
https://github.com/urbanopt/geojson-modelica-translator/blob/develop/geojson_modelica_translator/system_parameters/system_parameters.py#L148

Two options:
1. Remove all handling of default values, since they no longer exist in the sys-param file
1. Read default values from schema

vtnate:
A while ago we removed the 'defaults' section of sys-param files. This removes the code for reading those sections.
Refactor the `get_param` code in system_parameters.py to remove references to defaults which haven't existed for some time.
CI is sufficient
Resolves #495

### DIFF
--- a/geojson_modelica_translator/system_parameters/system_parameters.py
+++ b/geojson_modelica_translator/system_parameters/system_parameters.py
@@ -90,62 +90,34 @@ class SystemParameters:
                 new_path = Path(filepath) / match.value
                 parse(str(match.full_path)).update(self.param_template, new_path.as_posix())
 
-    # def resolve_defaults(self):
-    #     """This method will expand the default data blocks into all the subsequent custom sections. If the value is
-    #     specificed in the custom block then that will be used, otherwise the default value will be replaced"""
-    #     pass
-
-    def get_default(self, jsonpath, default=None):
-        """Return either the default in the system parameter file, or the specified default.
-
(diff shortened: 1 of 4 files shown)


geojson_modelica_translator/system_parameters/system_parameters.py
```diff
@@ -90,62 +90,34 @@ def resolve_paths(self):
...
-                results.append(default_value)
-            else:
-                results.append(match.value)
+        for match in matches:
+            results.append(match.value)
 
         if len(results) == 1:
-            # If only one value, then return that value and not a list of values
+            # If only one value, then return that value and not a list of values.
             results = results[0]
```
nllong, line 112: I don't like that we do this. I think we should always return a list. I don't know why I did this (I'm guilty for it). I think I was following another API pattern. Regardless, leave it as is, otherwise it will break lots of things.

geojson_modelica_translator/system_parameters/system_parameters.py
```diff
@@ -90,62 +90,34 @@ def resolve_paths(self):
...
         elif len(results) == 0:
-            return default_value
+            return None
```
nllong, line 114: nice, no results means NONE!

nllong (approved):

Issue #663, opened 2024-10-08
Branch off GHE and add dedicated circulation pump for GHE


JingWang-CUB:
### Expected Behavior
Each GHE should have its own branch connected to the main loop with a dedicated circulation pump. The controller determines whether the GHE branch is activated or not.

### Version Information

Modelica Buildings Library: `X.Y.Z`
Modelica Standard Library: `X.Y.Z`
TEASER: `X.Y.Z`
Python: `X.Y.Z`

JingWang-CUB:
This issue has been solved with PR #669 .

PR #669, opened 2024-10-22
Modify templates to support branched off configuration of GHEs

JingWang-CUB:
Now the GHEs are directly put on the ATL which does not comply with real-world implementations. 

Put GHEs on their dedicated branches and add separate pumps for them.
image

poetry run any 5G tests with GHE

#663


vtnate (approved):
Love seeing a more realistic representation!

PR #670, opened 2024-10-25
Support single building DES templates

nllong:
For the constant and variable DES systems, if there was a single building, then the model would fail to build.

If the jsonpath returns a dict, then it will assume it is a single building. This is not common, but needed for the aggregated load analysis.

unit tests should pass

n/a

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [tool.poetry]
 name = "geojson-modelica-translator"
 version = "0.8.0-rc1"
-description = "Package for converting GeoJSON to Modelica models for Urban Scale Analyses."
+description = "Package for converting GeoJSON to Modelica models for urban scale analyses."
 authors = ["URBANopt DES Team <nicholas.long@nrel.gov>"]
 license = "BSD-4-Clause"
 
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -36,7 +36,7 @@ geojson = "^3.1"
 jinja2 = "^3.1.2"
 jsonpath-ng = "~1.6"
 jsonschema = "~4.20"
-modelica-builder = "^0.5.1"
+modelica-builder = "^0.5.2rc3"
```
nllong, line 39: we should release 0.5.2 soon-ish

vtnate: I'd go so far as to say that if we merge my open PRs, we should go to 0.6.0. Agreed that should happen soon.

vtnate (commented):
The code looks good. I was expecting the test to check behavior when there is only a single building in the sys-param though, since handling that appears to be what the new code supports.

tests/system_parameters/test_system_parameters.py
```diff
@@ -192,6 +192,50 @@ def test_get_param(self):
         value = sp.get_param("not.a.real.path")
         assert value is None
 
+    def test_get_param_with_two(self):
```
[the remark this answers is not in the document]
vtnate: Previous test checks a single building. This returns a list, the desired behavior.

vtnate (approved):

Issue #675, opened 2024-10-29
Put GHE downstream of the buildings in the same building group


JingWang-CUB:
### Expected Behavior
To be consistent with how ThermalNetwork sizes GHEs (upstream method) and generates building groups, the GHEs in Modelica should be downstream of the buildings in the same group.

### Actual Behavior
GHEs are now put upstream of the buildings in Modelica.

### Version Information

Modelica Buildings Library: `X.Y.Z`
Modelica Standard Library: `X.Y.Z`
TEASER: `X.Y.Z`
Python: `X.Y.Z`

vtnate:
Closed by #677

Issue #666, opened 2024-10-15
5G ETS templates should take heat pump COP values from the sys_params file


JingWang-CUB:
### Expected Behavior
The 5G ETS templates should take the heat pump COP values from the user defined sys_params file. The COP value should be injected to somewhere near here in the building templates: [https://github.com/urbanopt/geojson-modelica-translator/blob/2ac21b7904aec6fb8dd33d21ef010ca651c4775f/geojson_modelica_translator/model_connectors/load_connectors/templates/TimeSeriesBuildingWithETS.mot#L18](url). 

### Version Information

Modelica Buildings Library: `X.Y.Z`
Modelica Standard Library: `X.Y.Z`
TEASER: `X.Y.Z`
Python: `X.Y.Z`

PR #677, opened 2024-11-19
Putting GHEs downstream of the buildings

JingWang-CUB:
This PR changes the couplings to put GHEs downstream of the buildings in the same building group.
Before:
image
After:
image

It also changes the time series building template to take heat pump heating COP from the sys_params file.

poetry run any tests with GHEs and buildings.

Issue #675 and issue #666.

### DIFF
--- a/geojson_modelica_translator/model_connectors/load_connectors/templates/TimeSeriesBuildingWithETS.mot
+++ b/geojson_modelica_translator/model_connectors/load_connectors/templates/TimeSeriesBuildingWithETS.mot
@@ -3,7 +3,8 @@ within {{ project_name }}.Loads.{{ model_name }};
   "Model of a building with an internal ETS. Building loads provided as time series"
   extends Buildings.DHC.Loads.Combined.BaseClasses.PartialBuildingWithETS(
     final allowFlowReversalSer=true,
-    {% endraw %}redeclare {{ project_name }}.Loads.{{ model_name }}.TimeSeriesBuilding bui(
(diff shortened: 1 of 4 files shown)


geojson_modelica_translator/model_connectors/load_connectors/templates/TimeSeriesBuildingWithETS.mot
```diff
@@ -3,7 +3,8 @@ within {{ project_name }}.Loads.{{ model_name }};
   "Model of a building with an internal ETS. Building loads provided as time series"
   extends Buildings.DHC.Loads.Combined.BaseClasses.PartialBuildingWithETS(
     final allowFlowReversalSer=true,
-    {% endraw %}redeclare {{ project_name }}.Loads.{{ model_name }}.TimeSeriesBuilding bui(
+    {% endraw %}COPHeaWat_nominal={{data['cop_heating']}},
```
vtnate, line 6: Does this `COPHeaWat_nominal` get used in this model, or another model? Was it being automatically set and now we're explicitly setting it, and the usage is already written?

JingWang-CUB: It gets used in every model with a 5G ETS model. We were using the MBL default values before and now we take the user-defined value from the sys_params file.

vtnate (approved):
🥳

PR #681, opened 2024-12-14
Allow skipping GeoJSON validation

nllong:
In some cases, the GeoJSON can contain extra information that is not problematic with the generation of the DES systems.

* allow skipping geojson validation. This should be considered an advance feature.

Code review should suffice


vtnate (approved):
kwargs and not an explicit skip_validation like you did in urbanopt_geojson?

LGTM 👍

nllong:
at some point we should fix the issue that we have two GeoJSON classes. One in the GMT and one in URBANopt-DES. I think the only logically approach is to have UO-DES use the version in GMT... so we'd have to move some functionality over to GMT. Thanks for reviewing this so quickly!

PR #682, opened 2024-12-16
Pre-commit and format markdown

nllong:
Small tweaks to the updated mkdocs migration!

* Add markdown linter and formatter
* fix spellings
* changelog -> change log
* e.g. | i.e. -> e.g., i.e,
* un-indent code blocks


geojson_modelica_translator/modelica/input_parser.py
```diff
@@ -155,7 +155,7 @@ def replace_within_string(self, new_string: str) -> None:
         """
         self.within = new_string
 
-    def find_model_object(self, obj_name: str) -> Tuple[Union[int, None], Union[str, None]]:
+    def find_model_object(self, obj_name: str) -> tuple[Union[int, None], Union[str, None]]:
```
vtnate, line 158: Since we now require Python 3.10, [type hints can be simpler](https://www.blog.pythonlibrary.org/2021/09/11/python-3-10-simplifies-unions-in-type-annotations/). We don't need to import `Union` any more, we can just do `tuple[int | None, str | None]`. I think (though I'm less confident) we also don't need to import `Any` now either.

Mostly meant as a reminder, not suggesting we tear apart the code looking for that!

.pre-commit-config.yaml
```diff
@@ -44,11 +44,22 @@ repos:
...
       # Run the linter
       - id: ruff
       #  unsafe fixes: https://docs.astral.sh/ruff/linter/#fixes
         args: [--fix, --exit-non-zero-on-fix, --output-format=full, --unsafe-fixes]
       # Run the formatter
       - id: ruff-format
+  - repo: https://github.com/igorshubovych/markdownlint-cli
+    rev: v0.43.0
+    hooks:
+      - id: markdownlint
```
vtnate, line 58: Linting in pre-commit is good for the soul 🫶

vtnate (approved):

PR #688, opened 2024-12-26
Remove `is_ghe_start_loop` from test geojson files

vtnate:

Removed `is_ghe_start_loop` from schema in https://github.com/urbanopt/urbanopt-geojson-gem/pull/280 because it is how being handled automagically [in ThermalNetwork](https://github.com/NREL/ThermalNetwork/pull/58).

- Remove this property from test geojson files

CI is sufficient


JingWang-CUB (approved):
Great that we now automatically handle this flag inside ThermalNetwork.

PR #689, opened 2025-01-01
Handle all-caps soil-temp stations, like Frankfurt

vtnate:
Some weather file names weren't being parsed correctly when looking up soil temp coefficients. Frankfurt was the first example we found of this. This handles those cases.

If soil temp coeff lookup fails, try using only the first word of the station and capitalizing it. If that still doesn't work, error as before.

Build a demo project with the uo-cli, using with the [Frankfurt weather file](https://energyplus.net/weather-location/europe_wmo_region_6/DEU/DEU_Frankfurt.am.Main.106370_IWEC). Go through the default baseline workflow, and `uo_des create-model`. If that works successfully (it did on my machine! 😄 ) we're good to go


JingWang-CUB (approved):
Looks good, thanks for fixing this so promptly!

PR #690, opened 2025-01-02
Clean up matching district generation types

### LINKED ISSUE #504: Be explicit about which generation we're working with in the code
### Expected Behavior
We now have models that will only work with a given generation of district system. We should arrange the file structure to make it clear which are shared and which belong to a specific district system.

Be explicit about which generation we're working with [here](https://github.com/urbanopt/geojson-modelica-translator/blob/develop/geojson_modelica_translator/model_connectors/couplings/coupling.py#L31-L34). Should be explicit `if` statements (and use `else` to capture unexpected items with a graceful error), or if/when we use minimum python 3.10 we could use a `case` statement.

We may want to improve the folder structure where models that only apply to a given generation of district are clearly separated out.

### Actual Behavior
4G is assumed, not explicitly stated, in the `else` statement

vtnate:
- Explicitly selects actions for 4th & 5th generation districts. Old technique was to assume that anything non-5th-gen was 4th gen. This enables additional generations of district systems to be modeled in the future, opening up development opportunities.
- DRY up gmt.py a bit.

CI is suffcient

Resolves #504

### DIFF
--- a/geojson_modelica_translator/model_connectors/couplings/coupling.py
+++ b/geojson_modelica_translator/model_connectors/couplings/coupling.py
@@ -28,10 +28,13 @@ class Coupling:
         self.district_type = district_type
         self._template_base_name = f"{model_a.model_name}_{model_b.model_name}"
 
-        if self.district_type == "fifth_generation":
-            self._template_dir = Path(__file__).parent / "5G_templates" / self._template_base_name
-        else:
-            self._template_dir = Path(__file__).parent / "templates" / self._template_base_name
+        match self.district_type:
+            case "fifth_generation":
+                self._template_dir = Path(__file__).parent / "5G_templates" / self._template_base_name
+            case "fourth_generation" | None:
+                self._template_dir = Path(__file__).parent / "templates" / self._template_base_name
+            case _:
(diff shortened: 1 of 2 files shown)


geojson_modelica_translator/model_connectors/couplings/coupling.py
```diff
@@ -28,10 +28,13 @@ def __init__(self, model_a, model_b, district_type=None):
         self.district_type = district_type
         self._template_base_name = f"{model_a.model_name}_{model_b.model_name}"
 
-        if self.district_type == "fifth_generation":
-            self._template_dir = Path(__file__).parent / "5G_templates" / self._template_base_name
-        else:
-            self._template_dir = Path(__file__).parent / "templates" / self._template_base_name
+        match self.district_type:
```
nllong, line 31: oh? Much better!

nllong (approved):

geojson_modelica_translator/model_connectors/couplings/coupling.py
vtnate: Ya, again courtesy of Python 3.10.

PR #691, opened 2025-01-02
Clean up loadbase

vtnate:
I had used a number of try/except blocks which was unnecessarily wordy. Cleaner!

- Replace some try/except blocks with dict.get()
- Remove an unnecessary `else: continue` from the if-statement
- Change print-statements to log items


nllong (approved):
sweet

PR #693, opened 2025-01-02
Bump dependency minimum versions

vtnate:
I looked through the dependency versions to see what could be updated with light/no effort. This is the result

- New versions of several dependencies.
    - BuildingsPy & PyTest had major version increments, without anything that breaks our workflow
    - Coveralls & PyTest-cov had changes that effectively dropped support for Python 3.13, so we're not changing those at this time. That will come at some point in the future.

CI is sufficient


vtnate:
@nllong Something in the upgrade to Poetry 2 broke the modelica formatter. Can you help me diagnose what's going on with that? The error message of `format` isn't very helpful 😆.

vtnate:
Huzzah, Poetry has a patch that fixes it for us!

PR #704, opened 2025-02-13
Remove coveralls

vtnate:
Something happened with coveralls that was returning a non-zero code. Decided we don't need coveralls, so removed it.

Removes `coveralls` from dependencies

CI is sufficient

No tickets, but it was causing CI failures on current PRs

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -390,27 +390,6 @@ tomli = {version = "*", optional = true, markers = "python_full_version <= \"3.1
 toml = ["tomli"]
 
 [[package]]
-name = "coveralls"
-version = "3.3.1"
-description = "Show coverage stats online via coveralls.io"
-optional = false
-python-versions = ">= 3.5"
-groups = ["dev"]
-markers = "python_version <= \"3.11\" or python_version >= \"3.12\""
-files = [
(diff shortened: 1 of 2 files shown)


tests/model_connectors/test_district_single_ghe.py
```diff
@@ -96,4 +96,7 @@ def test_simulate_district_system(self):
             f"{self.district._scaffold.project_name}.Districts.DistrictEnergySystem",
             run_path=self.district._scaffold.project_path,
             file_to_load=self.district._scaffold.package_path,
+            start_time=0,  # Day 0 (in seconds)
+            stop_time=3600,  # For 1 day duration (in seconds)
```
nllong, line 100:
```suggestion
            stop_time=86400,  # For 1 day duration (in seconds)
```

I think this is one hour... can we try 86400?

nllong (changes requested):
let's try one day!

tests/model_connectors/test_district_single_ghe.py
vtnate: My comment was incorrect, I really meant it to be one hour. Commit [34ca909](https://github.com/urbanopt/geojson-modelica-translator/pull/704/commits/34ca9093f8d97575ed27716e2cd9d2039e56badc) attempted one day of simulation and the CI runner failed. I have updated the comment.

nllong (approved):
looks good now!

PR #706, opened 2025-03-06
Expose skipping weather download

vtnate:
In cases where a user wants to run UO without an internet connection, they will need to provide their own epw & mos weather files. This not only exposes the existing functionality to the cli, we now also look in multiple common locations in the users' project folder for weather.

- Expose `skip_weather_download` to the cli
- Look for weather in multiple locations, not just the dir the sys-param file is in
- Update existing tests to work with the new weather location ideas

Run the cli by hand, with the new `-w` flag.
- Is `-w` a good short flag for weather? It's to skip downloading and use an existing file

I tried logging when we download and when we use a local file (which would have been useful for testing), and it wasn't showing on my stdout. Not sure why, so I am ignoring that for now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -113,6 +113,7 @@ tests/output
 tests/*/microgrid_output
 tests/output_ets
 tests/*/output
+tests/**/weather
 tests/modelica/data/packages/*/*.mat
 tests/modelica/data/packages/*/*.txt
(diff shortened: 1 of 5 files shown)


nllong (approved):
looks great... much more flexible!

PR #707, opened 2025-04-07
Enable selecting output variables from Modelica simulation

vtnate:
This enables selecting specific output variables for the Modelica simulation to emit, using [regex or wildcards](https://openmodelica.org/doc/OpenModelicaUsersGuide/1.24/omedit.html#output). I was also able to clean up the simulate.most templated script that we use to send the actual command to Modelica.

Enables new filtering functionality for the Modelica simulation

Run the `tests/management/test_uo_des.py::test_cli_runs_existing_5g_model_with_specific_variables` test and manually open the results file in OM to review the output. Confirm that the `PPum` variable is not available because that test does not specifically include it in the list of variables we are interested in.

### DIFF
--- a/geojson_modelica_translator/modelica/lib/runner/simulate.most
+++ b/geojson_modelica_translator/modelica/lib/runner/simulate.most
@@ -6,15 +6,6 @@ loadFile("{{ file_to_load }}");
 getErrorString();
 {% endif %}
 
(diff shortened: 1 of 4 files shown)


nllong (approved):
sweet, this worked as expected!

PR #708, opened 2025-04-08
Add autosizing flags to sys-param values that may be autosized by TN/GHED

vtnate:
Some values in the sys-param file get modified in-place by the [ThermalNetwork](https://github.com/nrel/thermalNetwork) package (which calls [GHED](https://github.com/BETSRG/GHEDesigner)). Being able to set these values explicitly instead of autosizing is sometimes valuable, so we need a way to tell TN/GHED not to operate on those values.

Adds new parameters to sys-param schema to notify whether to autosize values or not.


JingWang-CUB (approved):
Looks good to me!

tests/model_connectors/data/sdk_output_skeleton_13_buildings/run/baseline_scenario/ghe_dir/sys_params_proportional.json
```diff
@@ -587,14 +592,18 @@
...
-              "number_of_boreholes": 1633
+              "length_of_boreholes_autosized": true,
+              "number_of_boreholes": 1633,
+              "number_of_boreholes_autosized": true
             }
           }
         ]
       },
       "central_pump_parameters": {
+        "pump_design_head": 60000000,
```
mitchute, line 603: That's alotta Pascals 🦎

tests/system_parameters/data/system_params_ghe.json
```diff
@@ -109,10 +109,13 @@
       },
       "central_pump_parameters": {
         "pump_design_head": 60000,
-        "pump_flow_rate": 0.01
+        "pump_design_head_autosized": true,
+        "pump_flow_rate": 0.01,
+        "pump_flow_rate_autosized": true
```
mitchute, line 114: It might be good to have a test file that sets some of these to `false`.

geojson_modelica_translator/system_parameters/system_parameter_properties.json
```diff
@@ -1402,9 +1407,19 @@
...
+              "default": true
+            },
             "length_of_boreholes": {
               "description": "The length of the borehole (in meters), determined from Thermal Network sizing.",
               "type": "number"
+            },
+            "length_of_boreholes_autosized": {
+              "description": "True if the length of boreholes is autosized, false if set manually",
+              "type": "boolean",
+              "default": true
```
mitchute, line 1422: @JingWang-CUB we don't need to address this here, but we should coordinate on how the MBL GHE model will eventually accept user-defined GHE configurations. Presumably, this will be more than just them specifying nbh and length.

mitchute (commented):
One comment to consider regarding whether we need a "falsey" test.

tests/system_parameters/data/system_params_ghe.json
vtnate: Yes. I think that test should be in TN though, no? These "autosized" flags aren't used in the GMT.

tests/model_connectors/data/sdk_output_skeleton_13_buildings/run/baseline_scenario/ghe_dir/sys_params_proportional.json
vtnate: This is what happens when you have a software engineer try to make tests pass 😆. IIUC we can decrease the head pressure if we increase the flow rate. I can set the values to whatever is appropriate.

nllong (approved):
have these new params been propagated to the .mot files?

new params looks sweet and useful. I'm getting eager to write an optimization layer around this whole sweet of tools.

vtnate:
> have these new params been propagated to the .mot files?
These autosizing params are not used in the GMT but rather in ThermalNetwork & GHED (links in description).

PR #717, opened 2025-06-06
Create a dependabot config file

vtnate:
Dependabot seems like a useful tool for watching security vulnerabilities in dependencies. However, without any configuration, it was a bit of a hassle and we ignored it. This is the config file SEED uses for dependabot, figured we should try it here.

Create a config file for dependabot

If you have admin rights to this repo, check the security settings for dependabot.


nllong:
Looks swell

nllong (approved):

PR #725, opened 2025-06-10
Dependabot updates

vtnate:
Grouping Dependabot updates will hopefully reduce the number of PRs Dependabot submits.

This will not upgrade `rpds-py` because of the semantic versioning. We require `^0.22.0`, which will only upgrade to `0.22.x` because it assumes (correctly, imho) the bump to 0.23.0 is a breaking change.

- Adds grouping to dependabot config
- Change poetry dependency specification to use carets `^` instead of compatible versioning `~=`
- Build the GMT with Poetry in CI, to ensure it works as expected


nllong (approved):
looks great... hopefully this "just works"™

PR #742, opened 2025-07-22
Update result processing library for GHPs

tanushree04:
- update code so it can read output patterns with building IDs appended
Running, geojson-modelica-translator/tests/geojson_modelica_translator/test_results_ghp.py


nllong (approved):
thanks for the updates! Sorry for taking so long to review.

PR #753, opened 2025-09-09
Schema fixes for ThermalNetwork

vtnate:
I accidentally required `property_boundary` and `no_go_zone_boundary` in the sys-param file, when they already exist in the geojson file and are used from there. They should not be in the sys-param file.

- Add new borefield sizing type: `rectangle_constrained`
- Remove the offending items from the schema and the test file.

CI is sufficient


nllong (approved):

chriswmackey (approved):
Thanks for getting this together, @vtnate . It all looks good to me!

PR #796, opened 2025-12-22
Add undisturbed soil temp validation

kflemin:
Added validation of the system parameters key: district_system.fifth_generation.soil.undisturbed_temp

- This is a required key, so a value will always be present in the sys params file or the file will fail schema validation (before the code added here)
- The default value in the schema is 15. There is also a lookup csv previously committed to the repo that lists temperature value per weather station.
-  If the value in the sys params file differs by a value greater than the "threshold", the user will get a warning that they may want to modify the undisturbed temp value in their sys params file.
- If a value can't be found in the lookup, the user will get a warning that their value couldn't be validated. For erroneous weather file names however, that error will be caught upstream from this check and the user will get an error that the weather file couldn't be found.
- The threshold is currently set at 0.5 degrees C.
- Example warning message: 
WARNING  geojson_modelica_translator.model_connectors.plants.borefield:borefield.py:149 Undisturbed soil temperature is set to 18.3 °C in system parameters, which differs from the lookup value of 10.0 °C for weather station 'Buffalo-Greater Buffalo Intl AP '. Consider updating the undisturbed soil temperature value in the system parameters file.


geojson_modelica_translator/model_connectors/plants/borefield.py
```diff
@@ -104,6 +150,9 @@ def to_modelica(self, scaffold):
                 self.system_parameters.get_param_by_id(self.ghe_id, "$.pre_designed_borefield.borehole_x_coordinates")
             )
 
+        # Validate undisturbed soil temperature (validates and warns - does not change values)
+        self.validate_undisturbed_soil_temperature(template_data["soil"]["initial_ground_temperature"])
```
JingWang-CUB: Do we need a test for this new function?

nllong:
> The test passed for me locally. My only suggestion is to rename it test_soil_temp.py. Thanks!

@kflemin did you (or do you want to) change the file name?

kflemin:
thank you @JingWang-CUB!  @nllong I just renamed the file. this should be set.

Issue #655, opened 2024-09-09
GHE dimensions sometimes give errors on shape when they shouldn't


vtnate:
### Expected Behavior

[This calculation](https://github.com/urbanopt/geojson-modelica-translator/blob/develop/geojson_modelica_translator/system_parameters/system_parameters.py#L805-L808) should always provide valid "discriminants" for any rectangle.

### Actual Behavior
Creating something close to square in the internal UI gives values that trigger the ValueError in the code above.

### Version Information

Modelica Buildings Library: `X.Y.Z`
Modelica Standard Library: `X.Y.Z`
TEASER: `X.Y.Z`
Python: `X.Y.Z`

vtnate:
A GHE in the geojson with`feature["properties"]["footprint_area"] = 10227` and `feature["properties"]["footprint_perimeter"] = 404` is an example of what causes this error.

PR #816, opened 2026-02-20
update calculate_dimensions function to handle small geometry errors

kflemin:
The GHE calculate_dimensions function was updated to handle small measurement errors (<1%) in near-square shaped features.

Tests were added for this function, including testing the specific edge case described in the issue.

#655


JingWang-CUB (approved):
The tests passed from my end. Thanks for fixing this issue!

Issue #716, opened 2025-06-04
Horizontal pipe length list order should be updated since now buildings are upstream of GHX


JingWang-CUB:
We forgot to update the horizontal pipe lengths after we moved the buildings upstream of the GHX.

PR #817, opened 2026-02-20
Add tests for pipe lengths order relative to loop order

kflemin:
Add tests to ensure that pipe length ordering updates with the loop order of the features

The functionality referenced in issue #716 appears to have been implemented already.  I did add some tests to ensure that it is working appropriately. 

The only place I could find where these pipe lengths are used is here:
https://github.com/urbanopt/geojson-modelica-translator/blob/develop/geojson_modelica_translator/model_connectors/networks/templates/DesignDataSeries_Instance.mopt#L22-L23

#716


JingWang-CUB (approved):
Looks good to me!

